PR #1, opened 2021-08-30
fix: correctly handle GraphQL rate limiting

johnletey:


src/core/impl/ContractInteractionsLoader.ts
```diff
@@ -100,11 +100,20 @@ export class ContractInteractionsLoader implements InteractionsLoader {
...
 
   private async getNextPage(variables: ReqVariables): Promise<GQLTransactionsResultInterface> {
-    const response = await this.arweave.api.post('graphql', {
+    let response = await this.arweave.api.post('graphql', {
       query: ContractInteractionsLoader.query,
       variables
     });
 
+    while (response.status === 403) {
+      await this.sleep(30 * 1000); // 30 seconds.
```
ppedziwiatr, line 109: please move `30 * 1000` to `private static readonly _30seconds = 30 * 1000`

src/core/impl/ContractInteractionsLoader.ts
```diff
@@ -120,4 +129,8 @@ export class ContractInteractionsLoader implements InteractionsLoader {
 
     return txs;
   }
+
+  private async sleep(ms: number) {
```
ppedziwiatr, line 133: please specify return type explicitly (probably `Promise<unknown>` ;-))

ppedziwiatr, line 133: this probably can be static.

src/core/impl/ContractInteractionsLoader.ts
ppedziwiatr, line 109: please add `logger.debug('GQL rate limiting, waiting ${ContractInteractionsLoader._30seconds}s before next try.');` here

ppedziwiatr (changes requested):

PR #2, opened 2021-08-31
chore: switch private to protected

johnletey:


ppedziwiatr:
this change shouldn't be necessary ;-)

PR #4, opened 2021-08-31
refactor: make BlockHeightSwCache methods async

ppedziwiatr:


ppedziwiatr:
"put" and "contains" should be probably also async..

PR #7, opened 2021-09-03
replaced console.debug with logger

cedriking:
Found a `console.debug` while trying to run the app without showing logs.


ppedziwiatr (approved):

ppedziwiatr:
Good catch, thanks :-)

PR #15, opened 2021-09-09
fix: restoring backwards compatibility with interction transactions t…

ppedziwiatr:
* restoring backwards compatibility with interction transactions that do not conform to the tag ordering rule

TODO: add more detailed description of WTF has happened here...

### DIFF
--- a/src/core/modules/impl/TagsParser.ts
+++ b/src/core/modules/impl/TagsParser.ts
@@ -1,0 +1,48 @@
+import { GQLEdgeInterface, GQLTagInterface, LoggerFactory, SmartWeaveTags } from '@smartweave';
+
+/**
+ * A class that is responsible for retrieving "input" tag from the interaction transaction.
+ * Two tags formats are allowed:
+ * 1. "multiple interactions in one tx" format - where "Input" tag MUST be next to the "Contract" tag
+ *    See more at https://github.com/ArweaveTeam/SmartWeave/pull/51
+ * 2. "traditional" format - one interaction per one transaction - where tags order does not matter.
+ *
+ * More on Discord: https://discord.com/channels/357957786904166400/756557551234973696/885388585023463424
+ */
+export class TagsParser {
+  private readonly logger = LoggerFactory.INST.create('TagsParser');
+
+  getInputTag(interactionTransaction: GQLEdgeInterface, contractTxId: string): GQLTagInterface {
+    // this is the part to retain compatibility with https://github.com/ArweaveTeam/SmartWeave/pull/51
+    if (TagsParser.hasMultipleInteractions(interactionTransaction)) {
+      this.logger.debug('Interaction transaction is using multiple input tx tag format.');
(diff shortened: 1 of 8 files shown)


src/core/modules/impl/TagsParser.ts
```diff
@@ -0,0 +1,48 @@
...
+      // the "old way" - i.e. tags ordering does not matter,
+      // if there is at most one "Contract" tag
+      // (note: there might no "Contract" tag as well).
+      // - so returning the first occurrence of "Input" tag.
+      return interactionTransaction.node.tags.find((tag) => tag.name === SmartWeaveTags.INPUT);
+    }
+  }
+
+  static hasMultipleInteractions(interactionTransaction): boolean {
+    return interactionTransaction.node.tags.filter((tag) => tag.name === SmartWeaveTags.CONTRACT_TX_ID).length > 1;
```
cedriking, line 46: Shouldn't this be `tag.name === 'Contract'` ?

src/core/modules/impl/TagsParser.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export class TagsParser {
+  private readonly logger = LoggerFactory.INST.create('TagsParser');
+
+  getInputTag(interactionTransaction: GQLEdgeInterface, contractTxId: string): GQLTagInterface {
+    // this is the part to retain compatibility with https://github.com/ArweaveTeam/SmartWeave/pull/51
+    if (TagsParser.hasMultipleInteractions(interactionTransaction)) {
+      this.logger.debug('Interaction transaction is using multiple input tx tag format.');
+      const contractTagIndex = interactionTransaction.node.tags.findIndex(
+        (tag) => tag.name === SmartWeaveTags.CONTRACT_TX_ID && tag.value === contractTxId
```
cedriking, line 20: Shouldn't this be:
```js
interactionTransaction.node.tags.findIndex((tag) => tag.name === 'Contract' && tag.value === contractTxId);
```

Unless `SmartWeaveTags.CONTRACT_TX_ID === 'Contract'` This might be a bit confusing if that's the case?

src/core/modules/impl/TagsParser.ts
cedriking: Confirmed `CONTRACT_TX_ID` is equal to `'Contract'`.

src/core/modules/impl/TagsParser.ts
cedriking: Confirmed `CONTRACT_TX_ID` is equal to `'Contract'`.

src/core/modules/impl/TagsParser.ts
ppedziwiatr: yup, it's an enum :-)

src/core/modules/impl/TagsParser.ts
ppedziwiatr: SmartWeaveTags is an enum created to keep all the tags defined and used by the protocol in the one place.

PR #33, opened 2021-09-22
(WIP) feat: update cache before external contract call

ppedziwiatr:


src/__tests__/regression/test-cases.json
```diff
@@ -1,5 +1,4 @@
 [
-  "Daj-MNSnH55TDfxqC7v4eq0lKzVIwh98srUaWqyuZtY",
```
ppedziwiatr, line 2: restore.

Issue #51, opened 2021-11-02
refactor: state cache


ppedziwiatr:
State should be cached per block height _and_ per transaction id (instead of current caching per block height only - which is fine if you want to get "the last state for given block height", but not enough to load state for the exact transaction - which is required by both FCP optimizations and new writes between contracts).

Implementing this will allow to remove the "intermediary" cache introduced in https://github.com/redstone-finance/redstone-smartcontracts/issues/19 .

PR #52, opened 2021-11-04
refactor: state cache #51

ppedziwiatr:


ppedziwiatr:
remove `fcpOptimization` feature switch.

src/__tests__/regression/test-cases.json
```diff
@@ -34,7 +34,6 @@
   "dpRwKqXGO234CgHp5i_zr0Vv-caoej-EWd9cpQVcYhA",
   "4BM86e-zd_XBSY4nFFH3Kz_9GY2jkcEmnxe7CKqONIw",
   "QIa535saTE2QBSKwAdV1ChxoH--t4liW49twzZAW7JM",
-  "GzekjMfuPHIv0T95pjupchKeYEFVcqc-rrAJ0QgENR0",
```
ppedziwiatr, line 37: Described in https://discord.com/channels/786251205008949258/877628352297570314/907352942644985917
Basically this contract has a programming mistake in a way how balance is being checked (`balance[caller] < 1` without verifying if `caller` in the `balance` object). This in turn caused that different values were saved in state for different "deep copy" versions (`lodash.cloneDeep` vs `JSON.parse(JSON.stringify)` - the first was saving NaN, the second null, and since:
image

- this in turn caused the evaluated state to differ between these two versions.

ppedziwiatr:
YOLO!

PR #54, opened 2021-11-18
fix(): BlockHeightSwCache can not used to graphql fromHeight

zyjblockchain:
when use CacheableContractInteractionsLoader model run swc
can not use BlockHeightSwCache.cacheHeight to load next graphql fromHeight.
It is possible that the block height of the grapghql gateway is not equal to the block height of the current chain.
So I think we should use the height of the cache in the stateEvaluator.
This situation may cause graphql to request the same data repeatedly, but avoids the loss of data

### DIFF
--- a/src/plugins/CacheableContractInteractionsLoader.ts
+++ b/src/plugins/CacheableContractInteractionsLoader.ts
@@ -52,7 +52,7 @@ export class CacheableContractInteractionsLoader implements InteractionsLoader {
 
     const missingInteractions = await this.baseImplementation.load(
       contractId,
-      cachedHeight + 1,
+      fromBlockHeight,
       toBlockHeight,
       evaluationOptions
     );


ppedziwiatr:
I'm not sure I fully understand the issue here...especially the `It is possible that the block height of the grapghql gateway is not equal to the block height of the current chain` part - how is that possible? - can you please provide some concrete example? Does this issue affect the `arweave.net` gateway?

Calls to GQL are extremely slow and the solution proposed in this PR would slow down the state evaluation in a general-use-case scenario - as basically you're not using the cache at all - in fact - with the change from this PR - the `CacheableContractInteractionsLoader` makes no sense - as it effectively behaves exactly the same as the base, non-cached `ContractInteractionsLoader` - it always loads all the interactions.

If for some reason you don't want to use the `CacheableContractInteractionsLoader`, you can simply configure your `smartweave` instance with the non-cached `ContractInteractionsLoader`:

```ts
const smartweave = SmartWeaveNodeFactory.memCachedBased(arweave)
    .setInteractionsLoader(new ContractInteractionsLoader(arweave))
    .build();
```

zyjblockchain:
> 

yes, this way can solution my problem.
thanks

PR #85, opened 2022-01-18
Fix: `createTx()` default tag values no longer overwrite user-provided value

estill01:
Fix to issue:
https://github.com/redstone-finance/redstone-smartcontracts/issues/83#issue-1107230090


estill01:
Oops -- need to fix.

PR #87, opened 2022-01-21
feat: adding redstone-isomorphic-fetch

asiaziola:


src/utils/utils.ts
```diff
@@ -1,18 +1,11 @@
 /* eslint-disable */
 import cloneDeep from 'lodash/cloneDeep';
-
+import 'redstone-isomorphic-fetch';
```
ppedziwiatr, line 3: hmm...shouldn't this import be added in every place where `fetch` is being used? e.g. `ArweaveWrapper`?
Otherwise we're somewhat dependant on module's read (or bundling in case of webpack or other similar crap) order?

ppedziwiatr (changes requested):

PR #95, opened 2022-02-15
Ppe/the one that uses interaction streaming

ppedziwiatr:


ppedziwiatr:
I have noticed an issue with contract `LppT1p3wri4FCKzW5buohsjWxpJHC58_rgIO-rYTMB8` - with the streaming loader, it loads longer than with the default one.

On the other hand - loot contract or "standard" PSTs evaluate 20-25% faster.

Needs more investiagation / debugging.

PR #101, opened 2022-02-18
Ppe/the one with wasm support

ppedziwiatr:


src/__tests__/integration/wasm-deploy-write-read.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  it('should return exception for inf. loop function for dry run', async () => {
+    const result = await contract.dryWrite({
+      function: 'infLoop'
+    });
+
+    expect(result.type).toEqual("exception");
+    expect(result.errorMessage.startsWith("[RE:OOG")).toBeTruthy();
+  });
+
+  /*it('should skip interaction during contract state read if gas limit exceeded', async () => {
```
ppedziwiatr, line 129: for some reason this test throws "top-level" out of gas exception (whereas it should be caught in WasmContractHandlerApi and transformed to "exception" type result).
The exact same use case works fine though in a standalone script (contract-wasm.ts)

jest.config.js
```diff
@@ -24,7 +24,12 @@ module.exports = {
 
   testEnvironment: 'node',
 
+  "transformIgnorePatterns": [
```
ppedziwiatr, line 27: this dumb configuration is required to make assemblyscript/loader work with both jest and ts-node
https://github.com/AssemblyScript/assemblyscript/issues/2139#issuecomment-1046250511

src/core/modules/StateEvaluator.ts
```diff
@@ -94,6 +94,8 @@ export class DefaultEvaluationOptions implements EvaluationOptions {
   };
 
   sequencerAddress = 'https://gateway.redstone.finance/';
+
+  gasLimit = Number.MAX_SAFE_INTEGER;
```
ppedziwiatr, line 98: we're generous by default...

src/core/modules/impl/DefaultCreateContract.ts
```diff
@@ -16,10 +16,13 @@ export class DefaultCreateContract implements CreateContract {
...
     const srcTx = await this.arweave.createTransaction({ data: src }, wallet);
 
+    const contractType: ContractType = src instanceof Buffer ? 'wasm' : 'js';
+
     srcTx.addTag(SmartWeaveTags.APP_NAME, 'SmartWeaveContractSource');
     // TODO: version should be taken from the current package.json version.
     srcTx.addTag(SmartWeaveTags.APP_VERSION, '0.3.0');
-    srcTx.addTag('Content-Type', 'application/javascript');
+    srcTx.addTag(SmartWeaveTags.SDK, 'RedStone');
+    srcTx.addTag(SmartWeaveTags.CONTENT_TYPE, contractType == 'js' ? 'application/javascript' : 'application/wasm');
```
ppedziwiatr, line 25: contracts indexing in redstone-sw-gateway has to adopted to this..as the gateway keeps all contracts source in 'text' column 🤦

src/core/modules/impl/DefaultStateEvaluator.ts
```diff
@@ -183,6 +186,18 @@ export abstract class DefaultStateEvaluator implements StateEvaluator {
...
+        this.logger.debug('Interaction evaluation', singleInteractionBenchmark.elapsed());
+
+        interactionCall.update({
+          cacheHit: false,
+          intermediaryCacheHit: false,
+          outputState: stackTrace.saveState ? currentState : undefined,
+          executionTime: singleInteractionBenchmark.elapsed(true) as number,
+          valid: validity[interactionTx.id],
+          errorMessage: errorMessage,
+          gasUsed: result.gasUsed
```
ppedziwiatr, line 198: btw. - we will get a nice report of gas usage of each of the interaction.

src/core/modules/impl/wasmImports.ts
```diff
@@ -0,0 +1,83 @@
...
+    msg: {
+      "msg.sender": function () {
+        return wasmModule.exports.__newString(swGlobal.transaction.owner);
+      },
+    },
+    api: {
+      _readContractState: (fnIndex, contractTxIdPtr) => {
+        const contractTxId = wasmModule.exports.__getString(contractTxIdPtr);
+        const callbackFn = getFn(fnIndex);
+        console.log("Simulating read state of", contractTxId);
```
ppedziwiatr, line 57: there's probably no easy way to synchronize this call in AS...

src/legacy/smartweave-global.ts
```diff
@@ -74,6 +85,18 @@ export class SmartWeaveGlobal {
         throw new Error('Not implemented');
       }
     };
+
+    this.useGas = this.useGas.bind(this);
+  }
+
+  useGas(gas: number) {
+    if (gas < 0) {
+      return;
```
ppedziwiatr, line 94: nah, should rather throw here.

src/core/modules/impl/DefaultCreateContract.ts
jakub-wojciechowski: We don't need to implement it now, but
1) It might be easier for devs to work on the source code and pass raw source code to the sdk abstracting the compilation step
2) It may also be easier for debugging, monitoring to have access to the original contract's source code, at least in one of the columns in our gateway

src/core/modules/impl/ContractDefinitionLoader.ts
```diff
@@ -50,6 +51,10 @@ export class ContractDefinitionLoader implements DefinitionLoader {
     this.logger.debug('Tags decoding', benchmark.elapsed());
     benchmark.reset();
 
+    const contractSrcTx = await this.arweaveWrapper.tx(contractSrcTxId);
+    const contractType: ContractType =
+      getTag(contractSrcTx, SmartWeaveTags.CONTENT_TYPE) == 'application/javascript' ? 'js' : 'wasm';
```
jakub-wojciechowski, line 56: Shouldn't we be more careful checking if CONTENT_TYPE == 'application/wasm');

src/core/modules/impl/wasmImports.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+    contract: {
+      "Contract.id": function () {
+        return wasmModule.exports.__newString(swGlobal.contract.id);
+      },
+      "Contract.owner": function () {
+        return wasmModule.exports.__newString(swGlobal.contract.owner);
+      },
+    },
+    msg: {
```
jakub-wojciechowski, line 48: For me it's confusing, how about using only Transaction.owner?

src/core/modules/impl/ContractDefinitionLoader.ts
ppedziwiatr: 👍

jakub-wojciechowski (approved):

ppedziwiatr:
YOLO!

PR #102, opened 2022-02-21
Wasm testing (WIP)

ppedziwiatr:


src/__tests__/integration/wasm-pst-regression.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  balances: Map<string, number>;
+}
+
+LoggerFactory.INST.logLevel('fatal');
+
+const WARP_PST = "KT45jaf8n9UwgkEareWxPgLJk4oMWpI5NODgYVIF1fY";
+
+const END_BLOCK = 863867;
+
+class MockDefinitionLoader implements DefinitionLoader {
```
ppedziwiatr, line 36: cool idea. Other solution - as we've discussed - would be to use original contract loader, but substitute transactions loader.
Anyway, I like this modularity feature of our SDK. What I don't like is the cumbersome configuration of the client (..memCachedBased().setInteractionsLoader().setDefinitionLoader().build(), etc.)
- do you have idea how to make it less verbose?

PR #103, opened 2022-02-23
More informative errors

angrymouse:


src/core/modules/impl/DefaultCreateContract.ts
```diff
@@ -71,7 +71,7 @@ export class DefaultCreateContract implements CreateContract {
         transfer
       });
     } else {
-      throw new Error('Unable to write Contract Source.');
+      throw new Error('Unable to write Contract Source.\n'+resource.statusText.msg);
```
ppedziwiatr, line 74: hmm. `resource`? did you mean `response`?
Also - the `statusText` is defined as a string (so there is no "msg" property):
```
post(transaction: Transaction | Buffer | string | object): Promise<{
        status: number;
        statusText: string;
        data: any;
    }>;
```

How about changing to 
```
`Unable to write Contract Source.\n${response?.statusText}`
```

src/core/modules/impl/DefaultCreateContract.ts
```diff
@@ -115,7 +115,7 @@ export class DefaultCreateContract implements CreateContract {
     if (response.status === 200 || response.status === 208) {
       return contractTX.id;
     } else {
-      throw new Error('Unable to write Contract Initial State');
+      throw new Error('Unable to write Contract Initial State\n'+resource.statusText.msg);
```
ppedziwiatr, line 118: same here...

ppedziwiatr (changes requested):

src/core/modules/impl/DefaultCreateContract.ts
angrymouse: Yes, exactly

angrymouse (commented):
resource means response

PR #108, opened 2022-03-05
Ppe/updating cache on contract call

ppedziwiatr:


src/__tests__/regression/test-cases/read-state.json
```diff
@@ -1,7 +1,6 @@
 [
   "Daj-MNSnH55TDfxqC7v4eq0lKzVIwh98srUaWqyuZtY",
   "t9T7DIOGxx4VWXoCEeYYarFYeERTpWIC1V3y-BPZgKE",
-  "5pSyVjFI07z8mbLeQhYBMsQ4M_MPidXIGX6T77rnF2s",
```
ppedziwiatr, line 4: removing, as these 2 contracts are not compatible with rgw (https://github.com/redstone-finance/redstone-sw-gateway/issues/17#issuecomment-995899195)

PR #117, opened 2022-03-11
feat: add support for wasm contracts written in Go #116

ppedziwiatr:


src/core/modules/impl/wasm/go-wasm-imports.ts
```diff
@@ -0,0 +1,500 @@
...
+        this._callbackTimeouts = new Map();
+        this._nextCallbackTimeoutID = 1;
+
+        let go = this;
+        // it is safe to redeclare this for each new module in the global scope
+        // - this function is called only during module initialization.
+        globalJsModule.redstone.go = {
+            WasmModule: {
+                registerWasmModule: function (moduleId) {
+                    go._id = moduleId;
```
ppedziwiatr, line 61: Go doest not export its wasm function in wasm module `export` field. Instead - it attaches it to a global js object...
It also doesn't use the `module.imports` for the functions imported from js. Instead it loads them from the global js object.

To prevent any naming collisions (as there might be multiple Go contracts involved in the state evaluation - eg. if contract makes foreign calls):
1. Go wasm module generates unique module id (https://github.com/redstone-finance/redstone-contracts-wasm/blob/main/go/common/wasm.go#L31)
2. Go wasm module attaches all exported functions in the global js object - under object with key from point 1. (https://github.com/redstone-finance/redstone-contracts-wasm/blob/main/go/common/wasm.go#L34)
3. Go wasm module notifies the js host about the module id (https://github.com/redstone-finance/redstone-contracts-wasm/blob/main/go/common/wasm.go#L34)
4. js host - the "Go" class - moves the functions from the `global[moduleId]` to `Go.exports` field
5. All the "imports" are attached to `global.redstone.go[moduleId].imports` object - and then loaded from this field inside Go wasm module (eg. https://github.com/redstone-finance/redstone-contracts-wasm/blob/main/go/common/imports/global.go#L8)

PR #120, opened 2022-03-16
feat: vm2 PoC

ppedziwiatr:


src/__tests__/regression/test-cases/read-state.json
```diff
@@ -11,7 +11,6 @@
   "c25-RdheC6khcACLv23-XXg1W7YuA-VSZ_1_qnNFbhw",
   "ZT-70ovBlkF6cRIqvyHy5lC2LcjudsmCz9z19M4_QC4",
   "o-qJmQ4B0d6TnyA_awjhiBdiq0O4Vt_dNWU3pTnhTu8",
-  "BYRf00nIE4pGkJ8GyUY2PIJ1rBtxFPoWIu-0dd8iukg",
```
ppedziwiatr, line 14: this contract makes a lot of unsafe calls with loading transactions from Arweave - which recently (because of issues with Arweave gateway) randomly fail

PR #126, opened 2022-04-04
feat: internal writes for Rust

ppedziwiatr:


src/contract/HandlerBasedContract.ts
```diff
@@ -619,6 +619,8 @@ export class HandlerBasedContract<State> implements Contract<State> {
       caller: this._parentContract.txId()
     };
 
+    console.log('Contract caller', interaction.caller);
```
ppedziwiatr, line 622: remove

src/core/modules/impl/ContractHandlerApi.ts
```diff
@@ -47,6 +45,7 @@ export class ContractHandlerApi<State> implements HandlerApi<State> {
 
       const stateCopy = deepCopy(currentResult.state, executionContext.evaluationOptions.useFastCopy);
       this.swGlobal._activeTx = interactionTx;
+      this.swGlobal.caller = interaction.caller; // either contract tx id (for internal writes) or transaction.owner
```
ppedziwiatr, line 48: worst case we will restore the action.caller - though I prefer to use it this way.

PR #131, opened 2022-04-23
feat: allow to create unsigned interaction and post it separately

noomly:
This PR shouldn't include any breaking changes. It'll allow users to obtain an unsigned `Transaction` object of a write interaction in order to sign it manually and then post it as a bundle or simply through arweave.net as a second step.

To obtain said unsigned `Transaction`, I replaced `createInteraction` with `createUnsignedInteraction` and made it public. I also had to write a `createUnsignedTx` function in `src/legacy/create-tx.ts` that the original `createTx` now uses. The user can use this method, sign the returned object and pass it to `writeInteractionTx` or `bundleInteractionTx`.

I adapted `writeInteraction` and `bundleInteraction` to the new `createUnsignedInteraction`. They now call it and then sign it with `arweave.transactions.sign(tx, this.wallet)`.

I haven't tested it yet, I will report back on monday for this. The modifications being light it should be ok but we never know!

### DIFF
--- a/src/contract/HandlerBasedContract.ts
+++ b/src/contract/HandlerBasedContract.ts
@@ -9,6 +9,7 @@ import {
   ContractInteraction,
   createDummyTx,
   createTx,
+  createUnsignedTx,
   CurrentTx,
   DefaultEvaluationOptions,
   emptyTransfer,
@@ -34,6 +35,7 @@ import { TransactionStatusResponse } from 'arweave/node/transactions';
 import { NetworkInfoInterface } from 'arweave/node/network';
 import stringify from 'safe-stable-stringify';
 import * as crypto from 'crypto';
+import Transaction from 'arweave/node/lib/transaction';
 
 /**
  * An implementation of {@link Contract} that is backwards compatible with current style
@@ -205,10 +207,16 @@ export class HandlerBasedContract<State> implements Contract<State> {
     if (!this.wallet) {
       throw new Error("Wallet not connected. Use 'connect' method first.");
     }
-    const { arweave } = this.smartweave;
 
-    const interactionTx = await this.createInteraction(input, tags, transfer, strict);
-    const response = await arweave.transactions.post(interactionTx);
+    const interactionTx = await this.createUnsignedInteraction(input, tags, transfer, strict);
+    await this.smartweave.arweave.transactions.sign(interactionTx, this.wallet);
+
+    return this.writeInteractionTx(interactionTx);
+  }
+
+  async writeInteractionTx(interactionTx: Transaction): Promise<string | null> {
+    this.logger.info('Write interaction tx', interactionTx.id);
+    const response = await this.smartweave.arweave.transactions.post(interactionTx);
 
     if (response.status !== 200) {
       this.logger.error('Error while posting transaction', response);
@@ -229,8 +237,14 @@ export class HandlerBasedContract<State> implements Contract<State> {
     if (!this.wallet) {
       throw new Error("Wallet not connected. Use 'connect' method first.");
     }
-    const interactionTx = await this.createInteraction(input, tags, emptyTransfer, strict);
+    const interactionTx = await this.createUnsignedInteraction(input, tags, emptyTransfer, strict);
+    await this.smartweave.arweave.transactions.sign(interactionTx, this.wallet);
 
+    return this.bundleInteractionTx(interactionTx);
+  }
+
+  async bundleInteractionTx(interactionTx: Transaction): Promise<any> {
+    this.logger.info('Bundle interaction tx', interactionTx.id);
     const response = await fetch(`${this._evaluationOptions.bundlerAddress}gateway/sequencer/register`, {
       method: 'POST',
       body: JSON.stringify(interactionTx),
@@ -258,12 +272,13 @@ export class HandlerBasedContract<State> implements Contract<State> {
     };
   }
 
-  private async createInteraction<Input>(
+  async createUnsignedInteraction<Input>(
     input: Input,
     tags: { name: string; value: string }[],
     transfer: ArTransfer,
     strict: boolean
-  ) {
+  ): Promise<Transaction> {
+    this.logger.info('Unsigned interaction input', input);
     if (this._evaluationOptions.internalWrites) {
       // Call contract and verify if there are any internal writes:
       // 1. Evaluate current contract state
@@ -297,9 +312,8 @@ export class HandlerBasedContract<State> implements Contract<State> {
       }
     }
 
-    const interactionTx = await createTx(
+    const interactionTx = await createUnsignedTx(
       this.smartweave.arweave,
-      this.wallet,
       this._contractTxId,
       input,
       tags,
--- a/src/contract/HandlerBasedContract.ts
+++ b/src/contract/HandlerBasedContract.ts
@@ -215,7 +215,7 @@ export class HandlerBasedContract<State> implements Contract<State> {
   }
 
   async writeInteractionTx(interactionTx: Transaction): Promise<string | null> {
-    this.smartweave.arweave.transactions.verify(interactionTx);
+    await this.smartweave.arweave.transactions.verify(interactionTx);
 
     this.logger.info('Write interaction tx', interactionTx.id);
     const response = await this.smartweave.arweave.transactions.post(interactionTx);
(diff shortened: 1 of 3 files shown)


src/contract/HandlerBasedContract.ts
```diff
@@ -205,10 +207,16 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
 
-    const interactionTx = await this.createInteraction(input, tags, transfer, strict);
-    const response = await arweave.transactions.post(interactionTx);
+    const interactionTx = await this.createUnsignedInteraction(input, tags, transfer, strict);
+    await this.smartweave.arweave.transactions.sign(interactionTx, this.wallet);
+
+    return this.writeInteractionTx(interactionTx);
+  }
+
+  async writeInteractionTx(interactionTx: Transaction): Promise<string | null> {
```
ppedziwiatr, line 217: I believe all the public methods in this class should be reflected in the `Contract` interface - that this class is implementing

ppedziwiatr (changes requested):

src/contract/HandlerBasedContract.ts
noomly: done

noomly:
It seems there isn't a case where `bundleInteraction` would return `null` so I changed its signature (and `bundleInteractionTx`) accordingly. I also have two questions:

1. I realized that `createUnsignedTx` may not be as generically usable for a write vs a bundle as I thought as it seems that `bundleInteraction` doesn't accept any `ArTransfer` parameter. Should I make two different versions of `createUnsignedInteraction` for write and bundle?

2. I see that `bundleInteraction` returns `any`. Is this because the API will possibly change in the future? Or should I define a type: `{ bundlrResponse: response, originalTxId: interactionTx.id }` (in Contract.ts) and use it instead of any?

noomly:
I just saw while testing that there is an issue with the testing step (when `strict` is set to true) in `createUnsignedInteraction`. The problem comes from `callContract` as it calls `createTx` at line 584, which expects that a valid wallet is attached to the contract object or that an address is given as a parameter to `callContract`.

I think we should move this testing step to the posting methods `writeInteractionTx` and `bundleInteractionTx` as the transaction is expected to be signed there, so we could extract from it the actual `caller` (and the interaction input) and pass it to `callContract`.

However, `callContract` is also called at line 293. As this seems to be related to internal writes, I am not sure if it is safe to move it as well to the posting methods?

ppedziwiatr:
> I just saw while testing that there is an issue with the testing step (when `strict` is set to true) in `createUnsignedInteraction`. The problem comes from `callContract` as it calls `createTx` at line 584, which expects that a valid wallet is attached to the contract object or that an address is given as a parameter to `callContract`.
> 
> I think we should move this testing step to the posting methods `writeInteractionTx` and `bundleInteractionTx` as the transaction is expected to be signed there, so we could extract from it the actual `caller` (and the interaction input) and pass it to `callContract`.
> 
> However, `callContract` is also called at line 293. As this seems to be related to internal writes, I am not sure if it is safe to move it as well to the posting methods?

let me analyse this for a second...

src/contract/HandlerBasedContract.ts
```diff
@@ -224,13 +234,21 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
     }
-    const interactionTx = await this.createInteraction(input, tags, emptyTransfer, strict);
+    const interactionTx = await this.createUnsignedInteraction(input, tags, emptyTransfer, strict);
+    await this.smartweave.arweave.transactions.sign(interactionTx, this.wallet);
+
+    return this.bundleInteractionTx(interactionTx);
+  }
+
+  async bundleInteractionTx(interactionTx: Transaction): Promise<any> {
+    this.smartweave.arweave.transactions.verify(interactionTx);
```
ppedziwiatr, line 249: I believe this call is missing `await`

src/contract/HandlerBasedContract.ts
```diff
@@ -205,10 +207,18 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
-    const interactionTx = await this.createInteraction(input, tags, transfer, strict);
-    const response = await arweave.transactions.post(interactionTx);
+    const interactionTx = await this.createUnsignedInteraction(input, tags, transfer, strict);
+    await this.smartweave.arweave.transactions.sign(interactionTx, this.wallet);
+
+    return this.writeInteractionTx(interactionTx);
+  }
+
+  async writeInteractionTx(interactionTx: Transaction): Promise<string | null> {
+    this.smartweave.arweave.transactions.verify(interactionTx);
```
ppedziwiatr, line 218: and this one too

src/contract/HandlerBasedContract.ts
ppedziwiatr: also..the result doesn't seem to be used anywhere - I believe the idea was to throw if the `verify` return `false`?

ppedziwiatr:
> I just saw while testing that there is an issue with the testing step (when `strict` is set to true) in `createUnsignedInteraction`. The problem comes from `callContract` as it calls `createTx` at line 584, which expects that a valid wallet is attached to the contract object or that an address is given as a parameter to `callContract`.
> 
> I think we should move this testing step to the posting methods `writeInteractionTx` and `bundleInteractionTx` as the transaction is expected to be signed there, so we could extract from it the actual `caller` (and the interaction input) and pass it to `callContract`.
> 
> However, `callContract` is also called at line 293. As this seems to be related to internal writes, I am not sure if it is safe to move it as well to the posting methods?

Ok, so...
1. yes, the testing step must be moved to `writeInteractionTx` and `bundleInteractionTx` - and obv. extracted as a separate private method, to reduce c/p...

2. It is safe to move this to the posting methods as well - as it is only being used while posting interactions (basically it allows the SDK to determine whether the new transaction - that the user is trying to post - contains any internal writes calls)

noomly:
Actually it looks like it's not really safe to move the internal writes check to the post method as it's supposed to add tags to the transaction... Which invalidate the signature of said transaction.

Is the call to `callContract` actually necessary to check for internal writes? I'm not sure whether this call mutates the state of the contract instance. If it's not necessary I could leave the check at the transaction creation step without the call to `callContract`. And in that case, would it be ok to have the internal writes tags when evaluating the interaction when strict is true or should I strip these tags when running `callContract`?

noomly:
Sorry for the distraction....

I pushed the separation of the pre post checks (tx validity, strict mode and inner writes) to a separate method. The problem now being that we need to add the internal writes tags to the transaction before the user signs it

ppedziwiatr:
> Sorry for the distraction....
> 
> I pushed the separation of the pre post checks (tx validity, strict mode and inner writes) to a separate method. The problem now being that we need to add the internal writes tags to the transaction before the user signs it

n/p :-)

I'm thinking about an alternative implementation - would creating a type like
`export type SigningFunction = (tx: Transaction) => Promise<void>;`

and using it inside the `Contract.connect` function solve your use-case? 

The `Contract.connect` would be defined then as 
`connect(signer: ArWallet | SigningFunction): Contract<State>`
- to be backwards compatible.

The implementation would then would look like this:
```
connect(signer: ArWallet | SigningFunction): Contract<State> {
    if (typeof signer == 'function') {
      this.signer = signer;
    } else {
      this.signer = async (tx: Transaction) => {
        await this.smartweave.arweave.transactions.sign(tx, signer);
      };
    }
    return this;
  }
```

- so basically you would have an option to pass a custom signing function, that will obviously take the transaction as an argument.

noomly:
This looks like a way cleaner and simpler implementation! This should work perfectly for our case. I'm going to try this out asap, I'll keep you updated!

ppedziwiatr:
> This looks like a way cleaner and simpler implementation! This should work perfectly for our case. I'm going to try this out asap, I'll keep you updated!
> 
> Thank you Ppe :)

I've ammended a small fix and also added some dumb usage example - https://github.com/redstone-finance/redstone-smartcontracts/pull/133/files#diff-682e2d5055ac72630829270fe1574eff0858381c4cb431cba55944d516b30abbR31

ppedziwiatr:
Hey @noomly . Does the solution in https://github.com/redstone-finance/redstone-smartcontracts/pull/133 PR work for you?

ppedziwiatr:
@noomly , released in `0.4.77`

Issue #150, opened 2022-05-25
feat: store error message in the interaction result


ppedziwiatr:
..enhance the current validity report - instead of returning just the true/false - add a full error/exception message.

PR #151, opened 2022-05-25
fix: upToTransactionId fix; feat: store error message in the interact…

ppedziwiatr:
…ion result #150


src/core/modules/StateEvaluator.ts
```diff
@@ -77,6 +77,7 @@ export class EvalStateResult<State> {
   constructor(
     readonly state: State,
     readonly validity: Record<string, boolean>,
+    readonly errorMessages: Record<string, string>,
```
ppedziwiatr, line 80: adding as a separate fields, as we need to retain backward compatibility with how currently `validity` is being stored - e.g. some contracts, that read other's contract state, might currently use the validity object and make boolean comparisons.

PR #154, opened 2022-05-31
Ppe/the one with caching per transaction

ppedziwiatr:
TODO:

- [ ] rewrite migration guide
- [ ] optimize validity and errorMessages caching (one per contract, not per each entry)
- [x] fix warp instance creation in regression tests
- [x] remove lmdb cache implemention (was suppose to be 'blazingly fast' (3x times faster than leveldb), turned out to be 2x slower... 🤷‍♂️ )

This PR:
## 1. Changes how state is being cached. Instead of caching per block height, the state is now being cached per transaction. ##
The contracts' state cache is organised in this structure:
contractTxId -> sortKey -> cachedValue

CachedValue wraps the cached state, validity and error messages.
The sortKey is the value used to sort the transactions in the lexicographical order - https://github.com/warp-contracts/warp/blob/ppe/the-one-with-caching-per-transaction/src/core/modules/impl/LexicographicalInteractionsSorter.ts#L25

**The reason behind this change is that caching per block height is not enough if multiple interactions are at the same height and two contracts interact with each other**, e.g.:
image

In the above example, both the contract `C_A` and contract `C_B` have multiple interactions at block height `10`.
Additionally, the contract `C_A` makes a foreign read from contract `C_B` in transaction `C_A_10_2`.
Before the change from this PR (and also how original `smartweave.js` implementation would work), the state of the `C_B` would be read for the last interaction for this contract at height 10 - i.e.  the `C_B_10_3` transaction - which is wrong.

In order to correctly read the state of the `C_B` in this case:

> 1. the transactions from both the `C_A` and `C_B` at block height `10` should be lexicographically sorted (lower part of the image - the presented order is just an example)
> 2. The state for the `C_B` should be read for the first transaction, that in lexicographical order is before the `C_A_10_2` - i.e. the `C_B_10_2` transaction in this example (_NOT_ for the `C_B_10_3` transaction, as the old implementation would behave).

## 2. The file-based cache and knex-based cache have been removed. ##
The `BlockHeightSwCache` interface has been removed, instead the `SortKeyCache` https://github.com/warp-contracts/warp/blob/ppe/the-one-with-caching-per-transaction/src/cache/SortKeyCache.ts has been created.

The only cache implementation now available (aparat from simple in-memory key-value cached used for stuff like caching ContractDefinition) is the LevelDB cache https://github.com/warp-contracts/warp/blob/ppe/the-one-with-caching-per-transaction/src/cache/impl/LevelDbCache.ts
It is using https://www.npmjs.com/package/level library under the hood.
It should work in both node.js and browser (in browser the IndexedDB should be used).
There is also an option to create a non-persistent, in-memory only cache - with the help of https://github.com/Level/memory-level#memory-level
In order to use the in-memory version - set the `inMemory` flag to true in cache options - https://github.com/warp-contracts/warp/blob/ppe/the-one-with-caching-per-transaction/src/cache/impl/LevelDbCache.ts#L16

The LevelDB is a lexicographically sorted key-value database - so it's ideal for this use case - as it simplifies cache look-ups (e.g. lastly stored value or value "lower-or-equal" than given sortKey).
The cache for contracts are implemented as sublevels - https://www.npmjs.com/package/level#sublevel--dbsublevelname-options.

The default location for the node.js cache is `./cache/warp`.
The default name for the browser IndexedDB cache is `warp-cache`

In order to reduce the cache size, the oldest entries are automatically pruned.

## 3. The SmartWeave factory has been simplified. ##
- it now offers three main methods (one for arweave.net - like gateways with GQL endpoint, one for RedStone-like gateways and one usefull for writing integration tests with the ArLocal) - https://github.com/warp-contracts/warp/blob/ppe/the-one-with-caching-per-transaction/src/core/WarpFactory.ts#L58

If you want to have full control over the created instance - use https://github.com/warp-contracts/warp/blob/ppe/the-one-with-caching-per-transaction/src/core/WarpFactory.ts#L84

The old `SmartWeaveNodeFactory` and `SmartWeaveWebFactory` have been removed.

## 4. The SDK now caches the error messages. ## 
- this will allow to display errors from interactions in tools like SonAR (instead of showing only validity)

## 5. The calls to arweave.net/info have been completely removed. ##
If no concrete block height or sortKey is passed to readState or viewState method, the SDK simply loads all the interactions from the gateway (starting from the lastly cached interaction).
Because the behaviour of the Arweave gateway in case of passing null to min/max block height in the gql query params is unknown (https://discord.com/channels/908759493943394334/908766823342801007/983643012947144725) - the Arweave interactions loader filters all interactions, that have null block data.

## 6. Loads of refactors ## 
- e.g. the simplification of the `HandlerBasedContract.createExecutionContext` method,
simplification of the `CacheableStateEvaluator.eval`...

## 7. The caching for hook "onContractCall" has been temporarily removed ##
- as it needs to be fixed.
EDIT: it has been fixed now. The issue was that the state was cached for the current transaction (the one, that is currently being evaluated), not the previous one.

## 8. The workaround for no-caching of the Sequencer interactions from the last block has been removed. ##
- as we can now cache the state per exact transaction.
The workaround was needed when the per-block height cache was used, because it could have happen that after evaluating the state and caching the state at given block height - more interactions could be still registered on this block height by the sequencer


ppedziwiatr:
TODO:

- [ ] rewrite migration guide
- [ ] optimize validity and errorMessages caching (one per contract, not per each entry)

src/core/modules/impl/ArweaveGatewayInteractionsLoader.ts
```diff
@@ -123,14 +138,63 @@ export class ArweaveGatewayInteractionsLoader implements InteractionsLoader {
...
+          break;
+        }
+      }
+
+      this.logger.debug('Slicing:', {
+        fromIndex,
+        toIndex
+      });
+
+      sortedInteractions = sortedInteractions.slice(fromIndex || 0, toIndex || maxIndex + 1);
```
ppedziwiatr, line 182: `maxIndex + 1` looks really dumb :-)

src/core/modules/impl/ArweaveGatewayInteractionsLoader.ts
```diff
@@ -123,14 +138,63 @@ export class ArweaveGatewayInteractionsLoader implements InteractionsLoader {
...
+      to: toSortKey,
+      loaded: sortedInteractions.length,
       time: loadingBenchmark.elapsed()
     });
 
-    return interactions;
+    return sortedInteractions.map((i) => i.node);
+  }
+
+  private isLocalEnv() {
```
ppedziwiatr, line 195: no longer needed

src/core/modules/impl/CacheableStateEvaluator.ts
```diff
@@ -35,79 +34,61 @@ export class CacheableStateEvaluator extends DefaultStateEvaluator {
...
-          const index = missingInteractions.findIndex((tx) => tx.node.id === entry.interactionTxId);
-          if (index !== -1) {
-            this.cLogger.debug('Inf. Loop fix - removing interaction', {
-              height: missingInteractions[index].node.block.height,
-              contractTxId: entry.contractTxId,
-              interactionTxId: entry.interactionTxId
-            });
-            missingInteractions.splice(index);
-          }
+    // TODO: this is tricky part, needs proper description
```
ppedziwiatr, line 48: well, this is next thing to be refactored, but not in this PR...instead of keeping the track of called contracts/transactions in this weird array, maybe we could somehow use the fact, the we keep the whole Contracts tree (i.e. each contract that is being read by the "Smartweave.contracts.readContractState" has a "parent" contract set...

src/core/modules/impl/ContractHandlerApi.ts
```diff
@@ -195,7 +190,10 @@ export class ContractHandlerApi<State> implements HandlerApi<State> {
   private assignRefreshState(executionContext: ExecutionContext<State>) {
     this.swGlobal.contracts.refreshState = async () => {
       const stateEvaluator = executionContext.warp.stateEvaluator;
-      const result = await stateEvaluator.latestAvailableState(this.swGlobal.contract.id, this.swGlobal.block.height);
+      const result = await stateEvaluator.latestAvailableState(
```
ppedziwiatr, line 193: verify if the WasmHandlerApi has all the changes.

src/core/modules/impl/WarpGatewayInteractionsLoader.ts
```diff
@@ -81,34 +80,29 @@ export class WarpGatewayInteractionsLoader implements InteractionsLoader {
...
-        : `${this.baseUrl}/gateway/interactions`;
+      const url = `${this.baseUrl}/gateway/interactions-sort-key`;
 
       const response = await fetch(
         `${url}?${new URLSearchParams({
           contractId: contractId,
-          from: fromBlockHeight.toString(),
-          to: toBlockHeight.toString(),
+          ...(fromSortKey ? { from: fromSortKey } : ''),
+          ...(toSortKey ? { to: toSortKey } : ''),
```
ppedziwiatr, line 103: verify again if any changes are needed on the gateway side

bundle.js
```diff
@@ -30,9 +31,10 @@ const runBuild = async () => {
     bundle: true,
     outfile: './bundles/web.bundle.min.js',
     platform: 'browser',
-    target: ['es2020', 'chrome58', 'firefox57', 'safari11'],
+    target: ['esnext'],
     format: 'iife',
-    globalName: 'rsdk'
```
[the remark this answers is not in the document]
asiaziola: done

PR #165, opened 2022-06-20
FEAT: enable Atomic NFT with Warp deploy

twilson63:
Feature: Enable Atomic NFT functionality with Warp Deploy

```
As a developer     
I want to use Warp to deploy Atomic NFT contracts      
These contracts use the `data` property of a contract transaction to store a `digital asset`
So that I can combine the contract state with the digital asset making the contract and asset on in the same, and take advantage of the Warp SmartWeave execution engine.
```

This PR, contains a modification to the ContractData type allowing for an optional `data` property. The `data` property is an object that contains the following two properties:

* Content-Type - the content type of the asset
* Body - the data of the asset

``` js
const contractTxId = await warp.createContract.deploy({
  wallet,
  initState: initialState,
  data: { 'Content-Type': 'text/html', body: '<h1>HELLO WORLD</h1>' },
  src: contractSrc
});
```

Totally open to change in the API and approach, my reasoning for this approach, is to leverage typescript to guide the developer to remember to add the `Content-Type` property for the data asset, instead of the developer having to remember to add the `Content-Type` tag in an optional tags property. There may be better ways to do this.

The code is minimal and there is a test for the new, code, I can add more tests if this is the direction you want to deliver this feature. Also, if this is good, would you consider a special `AtomicNFT` code component, like you have for `PST` to make it even easier for developers to create and interact with `Atomic NFT` contracts.


ppedziwiatr:
I like the simplistic approach and minimal amount of changes in  the code :-)
I'll make some minor adjustments and merge it into main.

@asiaziola , can you please verify how the Warp gateway will behave with a change from this PR?

asiaziola:
> I like the simplistic approach and minimal amount of changes in the code :-) I'll make some minor adjustments and merge it into main.
> 
> @asiaziola , can you please verify how the Warp gateway will behave with a change from this PR?

you have green light, guys 🟢

ppedziwiatr:
hey @twilson63 , because of some weird commits history on this fork and conflicts with main repo, I've created a new branch on main repo and applied 281eb565a41a525d72e39d2f5c927a6cc0663d09 and 6b24f034f674895ebcbd78bb299192611a2aada8 commits: https://github.com/warp-contracts/warp/pull/166
I've also merged the test cases with the deploy-write-read.test.ts, as there was a lot of code duplication.

Your changes are available in `1.1.3` version.

Thanks again for the PR.

Re. dedicated API for AtomicNFT - we're more than happy to see another PRs ;-)

PR #179, opened 2022-07-01
Ppe/the one with api changes

ppedziwiatr:


src/contract/HandlerBasedContract.ts
```diff
@@ -178,56 +179,68 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
+        tags: effectiveTags,
+        strict: effectiveStrict,
+        vrf: effectiveVrf
+      });
+    } else {
+      const interactionTx = await this.createInteraction(input, effectiveTags, effectiveTransfer, effectiveStrict);
+      const response = await arweave.transactions.post(interactionTx);
+
+      if (response.status !== 200) {
+        this.logger.error('Error while posting transaction', response);
```
asiaziola, line 221: maybe let's log `response.status` and `response.statusText`

src/core/WarpFactory.ts
```diff
@@ -34,35 +34,83 @@ export const defaultCacheOptions: CacheOptions = {
...
+   * with Warp testnet (https://testnet.redstone.tools/)
+   */
+  static forTestnet(cacheOptions = defaultCacheOptions) {
+    const arweave = Arweave.init({
+      host: 'testnet.redstone.tools',
+      port: 443,
+      protocol: 'https'
     });
+
+    return this.customArweaveGw(arweave, defaultCacheOptions);
```
asiaziola, line 71: hmmm... not warpGw?

src/core/WarpFactory.ts
```diff
@@ -34,35 +34,83 @@ export const defaultCacheOptions: CacheOptions = {
...
+      return this.customWarpGw(arweave, defaultWarpGwOptions, cacheOptions);
+    }
   }
 
   /**
    * Returns a fully configured {@link Warp} that is using arweave.net compatible gateway
-   * (with a GQL endpoint) for loading the interactions.
    */
-  static arweaveGw(arweave: Arweave, cacheOptions: CacheOptions = defaultCacheOptions): Warp {
+  static customArweaveGw(arweave: Arweave, cacheOptions: CacheOptions = defaultCacheOptions): Warp {
```
asiaziola, line 106: i think we should set it as private as i dont really see the difference between this method and using forMainnet with some custom options but maybe we should leave it for the sake of symmetry with `customWarpGw` but maybe we should make both private as we've discussed today :>

jakub-wojciechowski: For the simplicity I will make only a single public static method "custom"

src/core/WarpFactory.ts
```diff
@@ -0,0 +1,136 @@
...
+ * All versions use the {@link Evolve} plugin.
+ */
+export class WarpFactory {
+  /**
+   * creates a Warp instance suitable for testing in a local environment
+   * (e.g. usually using ArLocal)
+   * @param arweave - an instance of Arweave
+   * @param cacheOptions - optional cache options. By default, the in-memory cache is used.
+   */
+  static forLocal(
```
jakub-wojciechowski, line 49: WDYT about making arweave an optional field and creating an instance if it's not passed to the factory?

src/contract/Contract.ts
```diff
@@ -22,10 +22,19 @@ interface BundlrResponse {
...
+  bundlrResponse?: BundlrResponse;
   originalTxId: string;
 }
+
+export type WriteInteractionOptions = {
+  tags?: Tags;
+  transfer?: ArTransfer;
+  strict?: boolean;
+  vrf?: boolean;
+  directWrite?: boolean;
```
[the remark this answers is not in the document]
jakub-wojciechowski: hmm, maybe "disableBundling" instead of directWrite?

Issue #152, opened 2022-05-28
vm2 making browser build fail


jfbeats:
Building using vite and using `import { SmartWeave, Contract } from 'redstone-smartweave'`

things from vm2 aren't affected by tree shaking so they are getting imported by rollup which fails to compile

```
 ERROR  [vite-plugin-pwa] 'VMScript' is not exported by __vite-browser-external, imported by node_modules/redstone-smartweave/lib/
esm/core/modules/impl/HandlerExecutorFactory.js
file: C:/Dev/ArweaveWebWallet/node_modules/redstone-smartweave/lib/esm/core/modules/impl/HandlerExecutorFactory.js:9:17
 7: import { Go } from './wasm/go-wasm-imports';
 8: import BigNumber from 'bignumber.js';
 9: import { NodeVM, VMScript } from 'vm2';
                     ^
10: class ContractError extends Error {
11:     constructor(message) {


 ERROR  error during build:                                                                                             10:13:18  
Error: 'VMScript' is not exported by __vite-browser-external, imported by node_modules/redstone-smartweave/lib/esm/core/modules/im
pl/HandlerExecutorFactory.js
    at error (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:198:30)
    at Module.error (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:12708:16)
    at Module.traceVariable (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:13067:29)
    at ModuleScope.findVariable (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:11711:39)
    at ChildScope.findVariable (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:6538:38)
    at ClassBodyScope.findVariable (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:6538:38)
    at ChildScope.findVariable (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:6538:38)
    at FunctionScope.findVariable (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:6538:38)
    at ChildScope.findVariable (C:\Dev\ArweaveWebWallet\node_modules\rollup\dist\shared\rollup.js:6538:38)
[...]

PR #190, opened 2022-07-10
bug: fixed bug when trying to import module using node esm and added …

twilson63:
…documentation to use ESM web bundles

Removed exports property from package.json, this property blocks the ability to consume the NodeJS ESM library using nodeJS ESM projects.

Also, I added some documentation on how to reference web-based ESM bundles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,14 +5,6 @@
   "main": "./lib/cjs/index.js",
   "module": "./lib/esm/index.js",
   "types": "./lib/types/index.d.ts",
-  "exports": {
-    ".": {
-      "import": "./lib/esm/index.js",
-      "require": "./lib/cjs/index.js"
-    },
-    "./esm": "bundles/esm.bundle.js",
-    "./web": "bundles/web.bundle.js"
-  },
   "sideEffects": false,
   "engines": {
     "node": ">=16.5"
--- a/README.md
+++ b/README.md
@@ -144,6 +144,24 @@ All exports are stored under `warp` global variable.
 </script>
 ```
 
+#### Using esm bundles
+
+Bundle files are possible to use in esm web environment only. Use minified version for production. It is possible to use latest or specified version.
+
+```html
+<!-- Latest -->
+<script src="https://unpkg.com/warp-contracts/bundles/esm.bundle.js"></script>
+
+<!-- Latest, minified-->
+<script src="https://unpkg.com/warp-contracts/bundles/esm.bundle.min.js"></script>
+
+<!-- Specific version -->
+<script src="https://unpkg.com/warp-contracts@1.0.0/bundles/esm.bundle.js"></script>
+
+<!-- Specific version, minified -->
+<script src="https://unpkg.com/warp-contracts@1.0.0/bundles/esm.bundle.min.js"></script>
+```
+
 ### Using the Warp Gateway
 
 #### SDK version >= `0.5.0`


package.json
```diff
@@ -5,14 +5,6 @@
   "main": "./lib/cjs/index.js",
   "module": "./lib/esm/index.js",
   "types": "./lib/types/index.d.ts",
-  "exports": {
-    ".": {
-      "import": "./lib/esm/index.js",
-      "require": "./lib/cjs/index.js"
-    },
-    "./esm": "bundles/esm.bundle.js",
-    "./web": "bundles/web.bundle.js"
```
ppedziwiatr, line 14: will this not break the https://github.com/warp-contracts/warp/issues/152#issuecomment-1142054586 ?

twilson63: To reference the bundled web and esm you would use 

```js

import { WarpWebFactory } from 'warp-contracts/bundles/esm.bundle'

```

instead of 

```js

import { WarpWebFactory } from 'warp-contracts/esm'

```

twilson63: FWIW, I continue to struggle with this issue, but I have narrowed down the root cause. The `./lib/esm/index.js` build is not esm compliant, I don't know why it is not, but I do know when I switch the exports config to point the `import` to cjs it works fine.

```json
"exports": {
    ".": {
      "import": "./lib/cjs/index.js",
      "require": "./lib/cjs/index.js"
    },
    "./esm": "bundles/esm.bundle.js",
    "./web": "bundles/web.bundle.js"
  },
```

So that may be another option. 

The current esm build results in this response:

```
file:///workspace/stamp-coin/copy-and-deploy.mjs:4
import { WarpFactory } from './warp/lib/esm/index.js'
         ^^^^^^^^^^^
SyntaxError: Named export 'WarpFactory' not found. The requested module './warp/lib/esm/index.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from './warp/lib/esm/index.js';
const { WarpFactory } = pkg;
```

For now, I am just pulling down the latest source and patching the package.json

```
npx degit warp-contracts/warp#v1.2.0-rc.1 warp
cd warp
npx json -I -f package.json -e 'this.exports.import = "./lib/cjs/index.js"'
yarn && yarn build:cjs
```

Please, let me know how I can help move this bug to fixed.

ppedziwiatr:
I believe that to make it compatible with .mjs files, we would need to add `"type": "module"` in the SDK's package.json...but this will probably break some other stuff :-)
The ESM version is built by the typescript compiler (https://github.com/warp-contracts/warp/blob/main/tsconfig.esm.json), so I really don't understand how can it be not compatible with esm module format :-)

We will return to this next week.

asiaziola:
I was able to make the .mjs file run correctly only using `tsup` npm package (https://github.com/egoist/tsup), looks cool and is based on `esbuild` but I'm not sure what's happening underneath. It requires a lot more tests but maybe it's a good start.

asiaziola:
@twilson63 @ppedziwiatr I was able to make the .mjs file run correctly only using tsup npm package (https://github.com/egoist/tsup), looks cool and is based on esbuild but I'm not sure what's happening underneath. It requires a lot more tests but maybe it's a good start.

ppedziwiatr:
Do we really need another weird library to run sth., that is considered as
a "standard"...?

Have you tried creating a very simple library and play with different
tsconfig settings?
Is `type: 'module'` the only way to make it work with mjs files?

niedz., 21 sie 2022, 15:56 użytkownik Asia ***@***.***>
napisał:

ppedziwiatr:
https://github.com/microsoft/TypeScript/issues/18442

This is insane...


niedz., 21 sie 2022, 16:32 użytkownik Piotr Pędziwiatr <
***@***.***> napisał:

ppedziwiatr:
https://github.com/microsoft/TypeScript/issues/18442#issuecomment-1037074273

niedz., 21 sie 2022, 16:37 użytkownik Piotr Pędziwiatr <
***@***.***> napisał:

ppedziwiatr:
Another 'interesting' solution
https://github.com/microsoft/TypeScript/issues/18442#issuecomment-1041545622

But this is all madness...I'm pretty sure that it will break sth in one of
those shitty web bundlers...

niedz., 21 sie 2022, 16:43 użytkownik Piotr Pędziwiatr <
***@***.***> napisał:

ppedziwiatr:
But I would start with the above. It's basically what we're doing (two
separate builds into cjs and esm format) with additional step of adding
package.json files into esm and cjs dists.

niedz., 21 sie 2022, 16:50 użytkownik Piotr Pędziwiatr <
***@***.***> napisał:

asiaziola:
Yeah, I was trying this one. I got 'Permission denied' when running sh script so I drop the idea 🤡  but as I see simply running 
```
chmod +x build.sh
```
should make it work. I'll try again and let you know.

```
git update-index --add --chmod=+x build.sh
```
should fix

asiaziola:
Weird af, .mjs works fine with `cjs` version, with `esm` I'm getting
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/asiaziola/arweave/redstone-smartcontracts/lib/esm/logging/index' imported from /Users/asiaziola/arweave/redstone-smartcontracts/lib/esm/index.js
```

asiaziola:
Seems like this is a cause https://bobbyhadz.com/blog/javascript-error-err-module-not-found-cannot-find-module#:~:text=The%20%22Error%20%5BERR_MODULE_NOT_FOUND%5D%3A,extension%20when%20importing%20local%20files.

asiaziola:
so I guess we'd need extensions for all the imported files?

ppedziwiatr:
And how does it work in the example that I've sent? There is also some
additonal configuration in the main package.json file...

niedz., 21 sie 2022, 17:16 użytkownik Asia ***@***.***>
napisał:

ppedziwiatr:
I.e. this part
https://github.com/jwalton/solox/blob/4d86f8779329ffc47552a16d2bab60544ee78c65/package.json#L6-L15

niedz., 21 sie 2022, 18:50 użytkownik Piotr Pędziwiatr <
***@***.***> napisał:

asiaziola:
I dont think we should compare these libs... they have one-file package. The problem is not what is written in package.json (we have it compatible with them), I believe the problem is how we import classes etc.

asiaziola:
I play with it some more :))))

ppedziwiatr:
I believe this has been finally fixed in `1.2.0-rc.2`, https://github.com/warp-contracts/bundlers#readme

Issue #213, opened 2022-08-29
internal writes errors handling


ppedziwiatr:
Raised by @jakub-wojciechowski during the Dex development (https://github.com/warp-contracts/wrc/tree/dex).
Currently errors have to be handled manually in the contract's code (checking the result.type from the `SmartWeave.contracts.write` and throwing ContractError if it is different then 'ok').

If `SmartWeave.contracts.write` fails, it should throw automatically by default.

ppedziwiatr:
done.

PR #214, opened 2022-08-29
feat: internal writes errors handling #213

ppedziwiatr:
wondering if the new `evaluationOptions.throwOnInternalWriteError` option should be true by default - as it might break backwards compatibility...or maybe we should set to `false` for some time (a month after release) and switch to `true` later?

### DIFF
--- a/src/__tests__/integration/internal-writes/internal-write-callee.test.ts
+++ b/src/__tests__/integration/internal-writes/internal-write-callee.test.ts
@@ -2,17 +2,17 @@
 import fs from 'fs';
 
 import ArLocal from 'arlocal';
-import Arweave from 'arweave';
-import { JWKInterface } from 'arweave/node/lib/wallet';
+import {JWKInterface} from 'arweave/node/lib/wallet';
 import path from 'path';
-import { mineBlock } from '../_helpers';
-import { Contract } from '../../../contract/Contract';
-import { Warp } from '../../../core/Warp';
-import { WarpFactory } from '../../../core/WarpFactory';
-import { LoggerFactory } from '../../../logging/LoggerFactory';
+import {mineBlock} from '../_helpers';
+import {Contract} from '../../../contract/Contract';
+import {Warp} from '../../../core/Warp';
+import {WarpFactory} from '../../../core/WarpFactory';
+import {LoggerFactory} from '../../../logging/LoggerFactory';
 
 interface ExampleContractState {
   counter: number;
+  errorCounter: number;
 }
 
 /**
@@ -49,13 +49,12 @@ describe('Testing internal writes', () => {
 
   let wallet: JWKInterface;
 
-  let arweave: Arweave;
   let arlocal: ArLocal;
   let warp: Warp;
   let calleeContract: Contract<ExampleContractState>;
   let calleeContractVM: Contract<ExampleContractState>;
-  let callingContract: Contract;
-  let callingContractVM: Contract;
+  let callingContract: Contract<ExampleContractState>;
+  let callingContractVM: Contract<ExampleContractState>;
   let calleeTxId;
   let callingTxId;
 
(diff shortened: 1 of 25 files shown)


src/core/modules/impl/handler/AbstractContractHandler.ts
```diff
@@ -36,17 +36,23 @@ export abstract class AbstractContractHandler<State> implements HandlerApi<State
   protected assignWrite(executionContext: ExecutionContext<State>, currentTx: CurrentTx[]) {
     this.swGlobal.contracts.write = async <Input = unknown>(
       contractTxId: string,
-      input: Input
+      input: Input,
+      throwOnError?: boolean
```
ppedziwiatr, line 40: @jakub-wojciechowski  - this is a new parameter that allow us to set the throwing behaviour independently from the evaluationOptions

src/__tests__/integration/internal-writes/internal-write-callee.test.ts
```diff
@@ -279,4 +278,100 @@ describe('Testing internal writes', () => {
       expect((await calleeContract2VM.readState()).cachedValue.state.counter).toEqual(634);
     });
   });
+
+  fdescribe('with internal writes throwing exceptions', () => {
```
ppedziwiatr, line 282: remove 'f'

src/__tests__/integration/internal-writes/internal-write-callee.test.ts
```diff
@@ -279,4 +278,100 @@ describe('Testing internal writes', () => {
...
+    beforeAll(async () => {
+      await deployContracts();
+    });
+
+    it('should auto throw on default settings', async () => {
```
ppedziwiatr, line 287: @jakub-wojciechowski here are new test cases:
1. 'default' settings - should auto throw
2. `evaluationOptions.throwOnInternalWriteError` set to `false` - should not auto throw
3. `evaluationOptions.throwOnInternalWriteError` set to `true` - should auto throw
4. `evaluationOptions.throwOnInternalWriteError` set to `true` but `throwOnError` in `SmartWeave.contracts.write` set to false - should not auto throw
5. `evaluationOptions.throwOnInternalWriteError` set to `false` but `throwOnError` in `SmartWeave.contracts.write` set to true - should throw

src/core/modules/StateEvaluator.ts
```diff
@@ -132,6 +132,8 @@ export class DefaultEvaluationOptions implements EvaluationOptions {
   walletBalanceUrl = 'http://nyc-1.dev.arweave.net:1984/';
 
   mineArLocalBlocks = true;
+
+  throwOnInternalWriteError = true;
```
ppedziwiatr, line 136: @jakub-wojciechowski the default should be set to `true`, right?

src/contract/testing/Testing.ts
```diff
@@ -9,12 +14,15 @@ export class Testing {
...
 
-  async generateWallet(): Promise<JWKInterface> {
+  async generateWallet(): Promise<Wallet> {
     this.validateEnv();
     const wallet = await this.arweave.wallets.generate();
     await this.addFunds(wallet);
 
-    return wallet;
+    return {
+      jwk: wallet,
```
ppedziwiatr, line 23: @jakub-wojciechowski additional change in the api that we've talked about.

src/__tests__/integration/internal-writes/internal-write-callee.test.ts
jakub-wojciechowski: Looks good, just double checking that default works exactly like `evaluationOptions.throwOnInternalWriteError` explicitly set to `true`, so when
6. `default` but `throwOnError` in `SmartWeave.contracts.write` set to `false` - should not auto throw

src/core/modules/StateEvaluator.ts
jakub-wojciechowski: Yes

src/core/modules/impl/handler/AbstractContractHandler.ts
jakub-wojciechowski: Great!

src/contract/testing/Testing.ts
jakub-wojciechowski: Awesome :)

src/__tests__/integration/internal-writes/internal-write-callee.test.ts
ppedziwiatr: it should behave like this, but I'll add a separate test case :-)

ppedziwiatr: test case added, works as expected.

PR #219, opened 2022-09-05
Fix compatibility with newer versions of wasm-bindgen (Rust contracts)

noomly:
There has been a change in the way `wasm-bindgen` generate the javascript glue code when compiling to WASM, which breaks an assumption Warp makes about its shape. Starting from the version `0.7.80` of wasm-bindgen, variables in the javascript glue code are now declared using `const` instead of `var`, which conflicts with the way Warp is extracting information from this code.

This PR fixes this compatibility issue by modifying the regex used to read the `dtor` value.


ppedziwiatr:
Thanks, gonna make some more checks, as the GH actions do not run on forks...

PR #231, opened 2022-09-28
fix: prevent wrapping contract Error

noomly:
Expose the actual contract error object for Rust contracts. This allows for precise contract error checking. See https://github.com/pianity/pianity-smartcontracts-next/blob/36ba4b8eb37f14f3f71ac8596094b61d2a30bebd/tests/src/fee.test.ts#L144 for a simple example of what it enables.

Note: it throws the exact error object that was returned by the contract. Not sure if we'd want to wrap it into a custom `Error` object?


ppedziwiatr:
I believe this one has been fixed by the @rpiszczatowski

PR #242, opened 2022-10-21
feat: the new testnet

ppedziwiatr:


ppedziwiatr:
oops, I've forgot about contracts deployment.

src/contract/deploy/impl/DefaultCreateContract.ts
```diff
@@ -23,7 +23,7 @@ export class DefaultCreateContract implements CreateContract {
 
     const source = new SourceImpl(this.arweave);
 
-    const srcTx = await source.save(contractData, wallet, effectiveUseBundler);
+    const srcTx = await source.save(contractData, this.warp.environment, wallet, effectiveUseBundler);
```
ppedziwiatr, line 26: warp instance should be passed in the constructor of the `SourceImpl`

src/core/WarpFactory.ts
```diff
@@ -67,14 +67,19 @@ export class WarpFactory {
    * with Warp testnet (https://testnet.redstone.tools/)
    */
   static forTestnet(
+    cacheOptions = defaultCacheOptions,
+    useArweaveGw = false,
     arweave = Arweave.init({
-      host: 'testnet.redstone.tools',
+      host: 'arweave.net',
```
asiaziola, line 73: `validateEnv` method in `Testing.ts` class will not work properly, we should probably check warp env

ppedziwiatr: you're right - it will now work only in local testing env.

but....since we're using real arweave now for the 'testnet'...the `addFunds` method would not work either way :/

not sure how to handle this - this is a general issue with using mainnet as testnet...

src/contract/testing/Testing.ts
```diff
@@ -1,39 +1,41 @@
...
-  private async addFunds(wallet: JWKInterface) {
+  async addFunds(wallet: JWKInterface) {
     const walletAddress = await this.arweave.wallets.getAddress(wallet);
     await this.arweave.api.get(`/mint/${walletAddress}/1000000000000000`);
   }
 
-  private validateEnv(): void {
-    if (this.environment !== 'local') {
+  async validateEnv(): Promise<void> {
+    if (!this.isArlocal) {
```
ppedziwiatr, line 23: `!(await this.isArlocal())`

src/core/Warp.ts
```diff
@@ -94,4 +93,17 @@ export class Warp {
 
     return this.plugins.get(type) as WarpPlugin<P, Q>;
   }
+
+  async generateWallet(): Promise<Wallet> {
+    const wallet = await this.arweave.wallets.generate();
+
+    if (this.testing.isArlocal()) {
```
ppedziwiatr, line 100: await

Issue #245, opened 2022-10-30
feat: allow http keep alive 


jarrvis:
For server side usage it would be nice to keep alive http connection.
E.g here:
https://github.com/warp-contracts/warp/blob/c1bfb2f58518a4369c044c84fabcb7d070d4e092/src/contract/HandlerBasedContract.ts#L293

or here:
https://github.com/warp-contracts/warp/blob/c1bfb2f58518a4369c044c84fabcb7d070d4e092/src/contract/deploy/impl/DefaultCreateContract.ts#L117
(since some users create contracts dynamically)

Is it possible with current api? Could we externalize fetch options to e.g. use custom configuration of http agent?

ppedziwiatr:
Hey, nope, it's not possible with current api. I'm thinking about adding an option to register in warp instance some function `(x: RequestInfo | URL) -> {}` - that would take the original request data as parameter and return object with options, that should be attached to the fetch.

We could probably use here the new plugin API - https://github.com/warp-contracts/warp/blob/rc/evm-signature/src/core/WarpPlugin.ts - the plugin would take `RequestInfo | URL` as an input and return `Partial<RequestInit>`.

It could be then registered with
```
warp.forMainnet().use(new FetchOptionsPlugin())
```

Last but not least - the SDK will need to verify whether such plugin was registered - and if so, call it to determine the options and attach them to fetch.

@asiaziola , such a great task for you!

jarrvis:
Hi @asiaziola, 👏👏👏
this is great!

For server side keep-alive usecase the plugin impl uses:
```javascript
fetchOptions = {
   agent: new http.Agent({
     keepAlive: true
   })
};
```

Debugging SSL shows:

HTTPS 97208: createConnection [Object: null prototype] {
  protocol: 'https:',
  slashes: true,
  auth: null,
  host: 'd1o5nlqr4okus2.cloudfront.net',
  port: 443,
  hostname: 'd1o5nlqr4okus2.cloudfront.net',
  hash: null,
  search: '?contractId=ECzhNQQBwAf3TV0r4rDyr6zglCzWha57VoFyNmqMnsQ&from=000001081630%2C1671580571027%2C3005273068d18d0699e446a9c12dfbc28a786c5012f2e1c8e6680423043c3efd&page=1&fromSdk=true&confirmationStatus=not_corrupted',
[...]

PR #255, opened 2022-11-14
feat: fetch options plugin

asiaziola:
solves https://github.com/warp-contracts/warp/issues/245

### DIFF
--- a/src/core/WarpFetchWrapper.ts
+++ b/src/core/WarpFetchWrapper.ts
@@ -1,0 +1,37 @@
+import { LoggerFactory } from '../logging/LoggerFactory';
+import { Warp } from './Warp';
+
+export interface FetchRequest {
+  input: RequestInfo | URL;
+  init: Partial<RequestInit>;
+}
+
+export class WarpFetchWrapper {
+  private readonly name = 'WarpFetchWrapper';
+  private readonly logger = LoggerFactory.INST.create(this.name);
+
+  constructor(private warp: Warp) {
+    this.warp = warp;
+  }
+
+  fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+    let fetchOptions: RequestInit;
+
+    if (this.warp.hasPlugin('fetch-options')) {
+      const fetchOptionsPlugin = this.warp.loadPlugin<FetchRequest, Partial<RequestInit>>('fetch-options');
+      try {
+        const updatedFetchOptions = fetchOptionsPlugin.process({ input, init: init || {} });
+        fetchOptions = { ...init, ...updatedFetchOptions };
+      } catch (e) {
+        if (e.message) {
+          this.logger.error(e.message);
+        }
+        throw new Error(`Unable to process fetch options: ${e.message}`);
+      }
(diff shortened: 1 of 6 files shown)


asiaziola:
classes where fetch has not been wrapped as it seems to not make much sense (or warp instance is not yet available): `Testing.ts`, `WarpGatewayContractDefinitionLoader.ts`, `WarpGatewayInteractionsLoader`, `smartweave-global.ts`, `ArweaveWrapper.ts`

ppedziwiatr:
> `WarpGatewayContractDefinitionLoader.ts`, `WarpGatewayInteractionsLoader`

imho it would make sense to add a wrapper here

src/core/WarpFetchWrapper.ts
```diff
@@ -0,0 +1,35 @@
+import { LoggerFactory } from '../logging/LoggerFactory';
+import { Warp } from './Warp';
+
+export interface FetchRequest {
+  input: RequestInfo | URL;
+  init: Partial<RequestInit>;
+}
+
+export class WarpFetchWrapper {
+  private readonly logger = LoggerFactory.INST.create('WarpFetchWrapper');
```
ppedziwiatr, line 10: I guess we should start using `WarpFetchWrapper.name` :-)

src/core/WarpFetchWrapper.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  constructor(private warp: Warp) {
+    this.warp = warp;
+  }
+
+  warpFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
```
ppedziwiatr, line 16: simply `fetch`? the class name already suggests that it is a wrapper...

tools/fetch-options-plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+import { defaultCacheOptions, WarpFactory } from '../src/core/WarpFactory';
+
+class FetchOptionsPlugin implements WarpPlugin<FetchRequest, RequestInit> {
+  process(request: FetchRequest): Partial<RequestInit> {
+    const url = request.input;
+
+    let fetchOptions = request.init;
+
+    if (url == `https://d1o5nlqr4okus2.cloudfront.net/gateway/sequencer/register`) {
+      fetchOptions = {
```
ppedziwiatr, line 16: hmm, maybe we should merge the options in the wrapper itself? Imho there's a risk that a plugin developer may forget to do a manual merge...

asiaziola: we can, it seemed easier for me to just update the object but maybe you're right.

ppedziwiatr: I believe it's more important to make it easier for developers to use than for us to implement ;-)

ppedziwiatr:
hmm, I would rebase, clean-up and merge this into main..hate those long-lasting PRs...

ppedziwiatr:
@asiaziola , how about a rebase and merge? or maybe ask Akord if they've had some time to test the new features...

PR #257, opened 2022-11-16
feat: deploy contract using evm signature

asiaziola:


src/contract/deploy/impl/DefaultCreateContract.ts
```diff
@@ -46,23 +57,24 @@ export class DefaultCreateContract implements CreateContract {
...
+    const { srcTxId, initState, tags, transfer, data } = contractData;
+    if (!this.wallet.signature) {
+      this.wallet.getSignature(contractData.wallet);
+    }
+    const signer = this.wallet.signature.signer;
 
     const effectiveUseBundler =
       disableBundling == undefined ? this.warp.definitionLoader.type() == 'warp' : !disableBundling;
 
-    let contractTX = await this.arweave.createTransaction({ data: data?.body || initState }, wallet);
```
asiaziola, line 54: not sure about this one - can we skip the wallet here reference here and just leave it when signing? seems to work.

ppedziwiatr: I believe all it does is https://github.com/ArweaveTeam/arweave-js/blob/master/src/common/common.ts#L98

But this is also done by the `sign` (https://github.com/ArweaveTeam/arweave-js/blob/master/src/common/transactions.ts#L231) method, so should be fine.

src/contract/deploy/impl/SourceImpl.ts
```diff
@@ -28,18 +28,31 @@ export interface SourceData {
...
     env: WarpEnvironment,
-    signer: ArWallet | SigningFunction,
+    wallet: ArWallet | Signature,
     useBundler = false
   ): Promise<any> {
     this.logger.debug('Creating new contract source');
 
     const { src, wasmSrcCodeDir, wasmGlueCode } = contractData;
 
+    this.wallet.getSignature(wallet);
```
ppedziwiatr, line 47: what's the point of this line?

src/contract/deploy/impl/SourceImpl.ts
```diff
@@ -28,18 +28,31 @@ export interface SourceData {
...
+    const signer = this.wallet.signature.signer;
+
+    if (this.wallet.signature.signatureType !== 'arweave' && !useBundler) {
```
ppedziwiatr, line 50: shouldn't this logic be extracted to the new Wallet class?

src/contract/Wallet.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class Wallet {
+  public signature: Signature;
+  private readonly warp: Warp;
+
+  constructor(warp: Warp) {
+    this.warp = warp;
+  }
+
+  getSignature(signature: ArWallet | Signature) {
```
ppedziwiatr, line 14: it would be awesome to add some tests to this wicked logic (especially since we now have it extracted to a separate class).

src/contract/deploy/impl/SourceImpl.ts
asiaziola: setting wallet property. User can use `save` method without deploying contract and the wallet instance should be then set from the begining. As an alternative I can probably pass the `wallet` of type Wallet in the deploy method and check in save method if wallet is of type Wallet or ArWallet | Signature. But it requires more 'ifs' and generally more complicated approach.

src/contract/deploy/impl/SourceImpl.ts
asiaziola: done.

Issue #244, opened 2022-10-29
feat: state cache entry pruning


JesseTheRobot:
This issue is for the implementation of the [mentioned cache pruning](https://github.com/warp-contracts/warp/blob/93350238248f0e65222923a6da08af19a4deb0e7/src/cache/impl/LevelDbCache.ts#L15) - I currently cannot locate the implementation for this feature - it's not a massively high priority, but it would be nice!
NB: I am willing to implement this myself, once I have free time (😅)

ppedziwiatr:
ah yeah, that comment is a bit misleading ;-) originally the SDK was updating the cache after each evaluated interaction, so it generated a lot of data...it has been largely optimised since then, but it can still add a lot of entries in cases where many internal writes are being used.

I implemented automatic the pruning at some point, but its implementation was not optimal at that time (I guess it cannot be done optimally in leveldb - cause we need to iterate over all keys) - the pruning was called for each 'put' into cache.

But it's good idea to add an option to prune manually.

Anyway, it should be fairly easy to implement, we're open for PRs

1. Add new method `SortKeyCache` interface - let's call it '`prune`. Implementation - `LevelDbCache`
2. The cache is organized into `sublevels`. Each `sublevel` is a separate contract.
3. use `const keys = await this.db.keys().all();` to get all sublevels (i.e. cached contracts). 
4. for each sublevel - (connect to sublevel via `const contractCache = this.db.sublevel<string, any>(contractTxId, { valueEncoding: 'json' })`) - call `sublevel.clear([options][, callback])` with `options` set to:
```
{
  lt: lastSortKey
}
```
The `lastSortKey` should be set to the result of the `SortKeyCache.getLast(contractTxId)`

ppedziwiatr:
Also - it might be worth considering adding a parameter to `prune` - `entriesStored` with a default value set to `1`.

ppedziwiatr:
Implemented in https://github.com/warp-contracts/warp/pull/268

PR #268, opened 2022-11-24
Implement prune method

janekolszak:


src/cache/impl/LevelDbCache.ts
```diff
@@ -136,7 +136,37 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+    return keys.length
+  }
+
   async prune(entriesStored = 5): Promise<PruneStats> {
-    throw new Error('Not implemented yet');
+    if (!entriesStored || entriesStored <= 0) {
+      entriesStored = 1
+    }
+
+    const keys = await this.db.keys().all();
```
ppedziwiatr, line 149: how about calling `this.allContracts` here? Instead of loading the contracts manually (which is exactly what `allContracts` is doing)

src/cache/impl/LevelDbCache.ts
```diff
@@ -136,7 +136,37 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+    }
+    return {
+      entriesBefore: 0,
+      sizeBefore: 0,
+      entriesAfter: 0,
+      sizeAfter: 0,
+    }
   }
-}
+}
```
ppedziwiatr, line 172: new line

src/cache/impl/LevelDbCache.ts
```diff
@@ -136,7 +136,37 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+
+      // Get keys that will be left, just to get the last one of them
+      const keys = await contractCache.keys({ reverse: true, limit: entriesStored }).all();
+      if (!keys || keys.length <= entriesStored) {
+        continue
+      }
+
+      contractCache.clear({ lt: keys[entriesStored - 1] })
+    }
+    return {
```
ppedziwiatr, line 165: please change the interface a bit so that the prune method would return `Promise<PruneState | null>`.

Null will simply mean that stats are not available.

Returning zeroes here might be misleading for the user (it may look as if the db was empty and nothing was removed;-))

ppedziwiatr (changes requested):

ppedziwiatr:
please modify commit message to: `feat: state cache entry pruning #244`

src/__tests__/unit/cache-leveldb-real-data.test.ts
```diff
@@ -0,0 +1,14 @@
+import { LevelDbCache } from '../../cache/impl/LevelDbCache';
+import { defaultCacheOptions, WarpFactory } from '../../core/WarpFactory';
+
+jest.setTimeout(90000)
+describe('Prune real Lmdb cache. Copy the real cache to ./dst, unskip and run yarn test:cache:real', () => {
```
ppedziwiatr, line 5: I'm not sure that this test should be run by default...the tests are already executing for quite long and the timeout above looks kinda scary ;-)

src/cache/impl/LevelDbCache.ts
```diff
@@ -136,7 +136,34 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+    for (let i = 0; i < contracts.length; i++) {
+
+      const contractCache = this.db.sublevel<string, any>(contracts[i], { valueEncoding: 'json' });
+
+      // manually opening to fix https://github.com/Level/level/issues/221
+      await contractCache.open();
+
+      // Get keys that will be left, just to get the last one of them
+      const entries = await contractCache.keys({ reverse: true, limit: entriesStored }).all();
+    //   console.log("A", contracts[i], entries, entries.length < entriesStored)
```
ppedziwiatr, line 159: please remove.

src/cache/impl/LevelDbCache.ts
```diff
@@ -136,7 +136,34 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+      if (!entries || entries.length < entriesStored) {
+        continue
+      }
+      await contractCache.clear({ lt: entries[entries.length - 1] })
```
ppedziwiatr, line 163: Can you please add this a comment to this method (so that other devs (and me, after a week ;-) could quickly understand what is going on here)

```
Let's assume that given contract cache contains these sortKeys: [a, b, c, d, e, f]
Let's assume entriesStored = 2
After pruning, the cache should be left with these keys: [e,f].

const entries = await contractCache.keys({ reverse: true, limit: entriesStored }).all();
This would return in this case entries [f, e] (notice the "reverse: true").

await contractCache.clear({ lt: entries[entries.length - 1] });
This effectively means: await contractCache.clear({ lt: e });
-> hence the entries [a,b,c,d] are removed and left are the [e,f]
```

ppedziwiatr (changes requested):

src/__tests__/unit/cache-leveldb-real-data.test.ts
ppedziwiatr: I guess `.skip` is missing here.

janekolszak: Ah, yes. Fixed.

src/__tests__/unit/cache-leveldb.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(await sut.getNumEntries()).toBe(contracts * 1)
+  });
+
+  it('should remove all unneeded entries, in many contracts', async () => {
+    const contracts = 100
+    const entriesPerContract = 20
+    const toLeave = 4
+    const sut = await cache(contracts, entriesPerContract)
+    await sut.prune(toLeave)
+    expect(await sut.getNumEntries()).toBe(contracts * toLeave)
```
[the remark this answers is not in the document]
janekolszak: Done. Checked if the cache contains newest elements and old are removed

ppedziwiatr (approved):

PR #269, opened 2022-11-24
Detektyw pikachu

ppedziwiatr:


src/core/modules/impl/handler/AbstractContractHandler.ts
```diff
@@ -54,6 +54,8 @@ export abstract class AbstractContractHandler<State> implements HandlerApi<State
 
       console.error('swGlobal.write call:', debugData);
 
+      // the calling contract current state should be put here into cache.
```
ppedziwiatr, line 57: this is now achieved by the 'uncommittedState'

PR #271, opened 2022-11-25
Replaced hadcoded gateway URL with gateway.redstone.finance 

SantiagoDePolonia:
gateway.redstone.finance domain is pointing to the same IP address.

### DIFF
--- a/src/core/WarpFactory.ts
+++ b/src/core/WarpFactory.ts
@@ -22,7 +22,7 @@ export type CacheOptions = {
   dbLocation: string;
 };
 
-export const WARP_GW_URL = 'https://d1o5nlqr4okus2.cloudfront.net';
+export const WARP_GW_URL = 'https://gateway.redstone.finance';
 
 export const defaultWarpGwOptions: GatewayOptions = {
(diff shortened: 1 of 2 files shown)


ppedziwiatr:
It's done on purpose.
We don't want to depend on DNS here, which is a single point of failure.

ppedziwiatr:
I.e. - "additional" DNS.

In fact, we should go even further here and switch directly to the ip address.

SantiagoDePolonia:
Personally I think that using additional DNS layer is a better architectural decision than using pure IP address. You can easier change where the domain in pointing without changing every hardcoded IP address. You can also easily implement additional proxy or some load balancer or some security layer. Thus you can scale it more easily.

ppedziwiatr:
We've already had a serious failure because of this DNS (sth related to configuration/renewal/payment issue, don't remember exactly) - that's why it was switched back (as previously the gateway.redstone.finance was used here) to a "direct" cloudfront address (which points to ALB either way, se we can still easily switch what is it pointing at).

Thanks for your suggestions, we will probably return to this at some point.

Issue #274, opened 2022-11-28
feat: add option to remove cached contract state


ppedziwiatr:
For both the LevelDb and Lmdb implementations.
Add method to the SortKeyCache interface - `delete(contractTxId: string)`

ppedziwiatr:
released in https://github.com/warp-contracts/warp/releases/tag/1.2.31

PR #281, opened 2022-11-30
Delete method

janekolszak:
Add delete method to the cache API, described in #274

### DIFF
--- a/src/cache/impl/LevelDbCache.ts
+++ b/src/cache/impl/LevelDbCache.ts
@@ -94,6 +94,12 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
     await contractCache.put(stateCacheKey.sortKey, value);
   }
 
+  async delete(contractTxId: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/cache/impl/LevelDbCache.ts
```diff
@@ -94,6 +94,12 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
     await contractCache.put(stateCacheKey.sortKey, value);
   }
 
+  async delete(contractTxId: string): Promise<void> {
+    const contractCache = this.db.sublevel<string, any>(contractTxId, { valueEncoding: 'json' });
```
ppedziwiatr, line 98: hmm..not sure how exactly this API works, but shouldn't rather sth like
`this.db.del(contractTxId)` be used here?
I.e. removing the whole contract from the 'root' database?

https://www.npmjs.com/package/level#dbdelkey-options-callback

janekolszak: I tried `this.db.del(...)` and it just doesn't work. I think this is because we're using sublevels.

ppedziwiatr (approved):

Issue #273, opened 2022-11-27
fix: the default LevelDB database should be initialized lazily


ppedziwiatr:
..instead of being initialized in the class constructor.

We now have an option to overwrite the default database (e.g. with the lmdb plugin https://github.com/warp-contracts/warp-contracts-lmdb) via the `useStateCache` and `useContractCache`.

Because the LevelDB cache is initialized in the constructor - the database files are created and the connection is opened - even if it is not effectively needed - because some other cache implementation is being used via `useStateCache` and `useContractCache`.

ppedziwiatr:
..or find some better method to overwrite the default cache - without complicating the WarpFactory api too much.

ppedziwiatr:
Released in https://github.com/warp-contracts/warp/releases/tag/1.2.31

PR #282, opened 2022-11-30
Lazy initialization of the LevelDB cache

janekolszak:
Lazy initialization upon first access, described in warp-contracts/warp#273


src/cache/impl/LevelDbCache.ts
```diff
@@ -22,19 +22,32 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+  private _db: MemoryLevel;
 
-  constructor(cacheOptions: CacheOptions) {
-    if (cacheOptions.inMemory) {
-      this.db = new MemoryLevel({ valueEncoding: 'json' });
+  // Configuration kept to initialize the cache upon first access
+  private readonly cacheOptions: CacheOptions;
+
+  // Lazy initialization upon first access
+  public get db(): MemoryLevel {
```
ppedziwiatr, line 31: I would set the getter as private - it should only used from within this class.

src/cache/impl/LevelDbCache.ts
```diff
@@ -22,19 +22,32 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+    if (!!this._db) {
```
ppedziwiatr, line 32: usually in such cases I prefer this flow:

```ts
if (!this._db) {
  // init shit
}

return this._db;
```

\- single return point from the function

ppedziwiatr (changes requested):

src/cache/impl/LevelDbCache.ts
```diff
@@ -22,19 +22,32 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+      return this._db
```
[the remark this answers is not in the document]
janekolszak: yes, sorry

src/cache/impl/LevelDbCache.ts
janekolszak: Changed, I was minimizing nested ifs

src/cache/impl/LevelDbCache.ts
```diff
@@ -22,19 +22,30 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+        this._db = new Level<string, any>(dbLocation, { valueEncoding: 'json' });
       }
-      const dbLocation = cacheOptions.dbLocation;
-      this.logger.info(`Using location ${dbLocation}`);
-      this.db = new Level<string, any>(dbLocation, { valueEncoding: 'json' });
     }
+    return this._db;
+  }
+
+  constructor(cacheOptions: CacheOptions) {
```
ppedziwiatr, line 47: since cache options is now a class field - `private readonly` can be added here, and line 28 and 48 removed.

ppedziwiatr (changes requested):

ppedziwiatr (approved):

Issue #284, opened 2022-12-02
perf: minimze communication between host and WASM module


ppedziwiatr:
especially reading/writing state from/to WASM module.

1. Fix double state initialization - https://github.com/warp-contracts/warp/blob/main/src/core/modules/impl/DefaultStateEvaluator.ts#L65-L79
2. Implement fix from https://github.com/warp-contracts/warp/blob/main/src/core/modules/impl/handler/WasmHandlerApi.ts#L41
3. Try to optimize executionContext creation - i.e. recreating the execution context constantly for the same contracts for consecutive internal writes.
4. block logging large objects from WASM module (or implement logging internally and only ask the host for the current logging level)

ppedziwiatr:
https://medium.com/@wl1508/avoiding-using-serde-and-deserde-in-rust-webassembly-c1e4640970ca - but that's unusable in our case.

ppedziwiatr:
One additional thing to consider is to pass binary data with the help of msgpack - which in general should be faster than json
https://news.ycombinator.com/item?id=20604597
https://msgpack.org/index.html - wide support in multiple langs, including AS, Rust, Go and obviously js.

msgpack is already used internally by the LMDB cache for the SDK (https://github.com/warp-contracts/warp-contracts-lmdb#warp-contracts-lmdb-cache)

First thing to check would be to change the implementation of the `currentState` in Rust - serialize the data in WASM using https://github.com/3Hren/msgpack-rust (https://docs.rs/rmp-serde/latest/rmp_serde/#examples or low-level - https://docs.rs/rmp/latest/rmp/#the-rust-messagepack-library) and deserialize in WasmHandlerApi (using either https://github.com/kriszyp/msgpackr#msgpackr or 
https://github.com/streamich/json-joy/blob/master/src/json-pack/README.md
- both claim to "blazingly fast", the first one is used natively by lmdb-js, the 2nd should work also in the browser).

I believe this would allow to pass data as buffers (between WASM and host).
Not sure if using js_sys.Uint8Array is the most efficient way (https://stackoverflow.com/questions/60171056/wasm-bindgen-arrays-of-u8-as-inputs-and-outputs-generated-javascript-has-diffe)

@noomly , I will obviously work on minimizing the calls between SDK and WASM, I wonder if you would like to experiment a little bit with passing the data using msgpack and buffers?

noomly:
[...]

PR #302, opened 2022-12-22
Allow using a different serialization format for the initial state and the Rust WASM<->JS bridge 

noomly:
Implements the idea discussed here https://github.com/warp-contracts/warp/issues/284#issuecomment-1359224307.

It allows the usage of alternative serialization formats to JSON for the initial state and the communication between WASM & JS. It implements Messagepack as a first alternative and is currently only focused on bringing this to Rust contracts but it lays the foundations for other formats & contracts.

This PR is still a WIP but I thought I'd submit a PR so that we can discuss the implementation details.


src/contract/deploy/impl/DefaultCreateContract.ts
```diff
@@ -85,9 +91,34 @@ export class DefaultCreateContract implements CreateContract {
     contractTX.addTag(SmartWeaveTags.SDK, 'RedStone');
     if (data) {
       contractTX.addTag(SmartWeaveTags.CONTENT_TYPE, data['Content-Type']);
-      contractTX.addTag(SmartWeaveTags.INIT_STATE, initState);
+      contractTX.addTag(SmartWeaveTags.INIT_STATE_FORMAT, stateFormat);
+      contractTX.addTag(
+        SmartWeaveTags.INIT_STATE,
+        typeof initState === 'string' ? initState : new TextDecoder().decode(initState)
```
noomly, line 97: I haven't tested it outside of Arlocal yet and I'm not sure how if Mainnet would appreciate a transaction having a binary tag value? Anyway it doesn't really seem right, it should probably be encoded in base64... What do you think? And if base64, should we use base64url or plain base64?

ppedziwiatr: I believe Arweave requires all tags to be encoded in Base64Url - https://github.com/ArweaveTeam/arweave-js/blob/master/src/common/lib/transaction.ts#L125

noomly: Oh that's true, completely forgot about that! So it's handled for us then, perfect!

ppedziwiatr:
I've briefly reviewed the PR and I really like the implementation. The only thing that bothers me is the c/p of the rust-wasm-imports..it will be a bit painful to maintain...

noomly:
Yeah I agree, there's definitely something to be done here. I'm still not sure what though; some functions implementations clash between the original json and the msgpack version... There are definitely some common parts that could be shared but honestly, it's probably still going to be a pain to maintain even if we try to merge the two rust-wasm-imports.

Maybe we could think of another strategy for the way this is handled (at least for Rust contracts, I haven't taken a close look at how it's handled for Go & AS). As I understand it, the only reason the glue code is reconstructed is to expose JS functions to WASM like the various SmartWeave functions & log, right? Have you considered trying to inject these functions by doing some light parsing on the glue code and leaving everything else intact for example?

ppedziwiatr:
1. I believe the Go and AS will be removed in the near future - there's exactly ZERO interest in the community in writing contracts in those language (plus both Go and AS have some technical issues in case of WASM); and we're simply unable to properly maintain all the versions.
2. I'm not sure what do you understand by "light parsing"...this whole mess (yes, I'm aware that it is a mess) with WASM glue code is caused by the fact, that wasm-bindgen mangles the function names in the wasm module with each compilation - I would have to first read (i.e. eval) the generated js with the glue code - which is kinda risky, cause I would have to trust some external code...
That's why this whole 'mapping' is done (the SDK runtime is 'static')
https://github.com/warp-contracts/warp/blob/main/src/core/modules/impl/wasm/rust-wasm-imports.ts#L287

But now that I read again the https://github.com/rustwasm/wasm-bindgen/issues/1128#issue-393781607 issue (which is the exact same issue we're facing in the SDK) - maybe there's some better way (e.g. https://github.com/rustwasm/wasm-bindgen/issues/1128#issuecomment-449820516).

We're open for PRs!

noomly:
1. Good to know! Less *mess* to maintain :)
2. I understand the security concern of using the glue code as-is; that discards my idea then.

Do you consider this issue of glue code handling critical for this PR or could it be researched and addressed in separate one at a later time?

ppedziwiatr:
ad. 2 - this should be done/refactored within a separate issue.

ppedziwiatr:
https://github.com/warp-contracts/warp/issues/309, @rpiszczatowski will try to refactor this crap :-)

ppedziwiatr:
@noomly , I believe we could merge this PR as is...the only thing missing is an integration test for the new format...
sth similar to https://github.com/warp-contracts/warp/blob/main/src/__tests__/integration/wasm/rust-deploy-write-read.test.ts

noomly:
hey @ppedziwiatr , I'll be back from vacation on the 9th, I'll be able to add the tests then.

ppedziwiatr:
> hey @ppedziwiatr , I'll be back from vacation on the 9th, I'll be able to add the tests then.

no rush :-)

ppedziwiatr:
btw. @noomly I've released the `1.2.48-beta.1` of the internal writes refactor (https://github.com/warp-contracts/warp/pull/311) - which is a prerequisite to a more advanced optimization for WASM (like reusing the handlers, instead of recreating them and initializing the internal wasm module state every time)

Feel free to test it out. I'll try to return to the https://github.com/warp-contracts/warp/pull/288 this week.

noomly:
Great! I'll get to this when I'm done with this PR :)

By the way, while running the typechecker I've realized that I made the `stateFormat` field of `CommonContractData` non-optional; which breaks calls to `warp.createContract.deploy` as they have to now include this new field. I feel like it would be better to have this field being explicitly set by the user as it'd depend on their contract's implementation; although it would be a breaking.

What do you think? Should we make it an optional field that defaults to JSON or should we leave it as a required field?

ppedziwiatr:
> What do you think? Should we make it an optional field that defaults to JSON or should we leave it as a required field?

We can't break backwards compatibility here (also - probably 99.9% of the users will still use the json), so please set the default value to JSON.

Thanks!

ppedziwiatr:
@rpiszczatowski we need to return to this :-)

PR #306, opened 2022-12-30
fix: reading owner of tx in strict mode

koonopek:
This MR solve problem where in strict mode, `customSignature` is not resolving to proper address.

It fails here: https://github.com/warp-contracts/warp/blob/main/src/contract/HandlerBasedContract.ts#L581 (thanks @asiaziola)
When we try to decode `address` from `signature` and we strictly assume that it is arweave `signature`. It only happens when `{stricit : true}`.

To solve this issue I extract owner assigned by `signature.signer` function and it solves the issues.

However, I also wanted to remove this code https://github.com/warp-contracts/warp/blob/main/src/contract/HandlerBasedContract.ts#L581, because of it - we are signing twice per one interaction (one dummy interaction just to get address and second real one). If interaction is signed by metaMask for example it will break the ux (two pop ups). However, when i removed this logic, it broke internal writes integration tests, so I left it there.

I have also tested it with - https://github.com/warp-contracts/warp-contracts-plugins/blob/main/warp-contracts-plugin-signature/src/server/evm/evmSignature.ts I didnt want to bring plugin as dependency so it is not in tests

### DIFF
--- a/src/contract/HandlerBasedContract.ts
+++ b/src/contract/HandlerBasedContract.ts
@@ -33,7 +33,7 @@ import {
 import { ArTransfer, ArWallet, emptyTransfer, Tags } from './deploy/CreateContract';
 import { InnerWritesEvaluator } from './InnerWritesEvaluator';
 import { generateMockVrf } from '../utils/vrf';
-import { Signature, SignatureType } from './Signature';
+import { Signature, CustomSignature } from './Signature';
 import { ContractDefinition } from '../core/ContractDefinition';
 import { EvaluationOptionsEvaluator } from './EvaluationOptionsEvaluator';
 import { WarpFetchWrapper } from '../core/WarpFetchWrapper';
@@ -366,13 +366,6 @@ export class HandlerBasedContract<State> implements Contract<State> {
       });
 
       this.logger.debug('Tags with inner calls', tags);
-    } else {
-      if (strict) {
-        const handlerResult = await this.callContract(input, undefined, undefined, tags, transfer, strict, vrf);
-        if (handlerResult.type !== 'ok') {
-          throw Error(`Cannot create interaction: ${handlerResult.errorMessage}`);
-        }
-      }
     }
 
     if (vrf) {
@@ -394,6 +387,14 @@ export class HandlerBasedContract<State> implements Contract<State> {
       this.warp.environment === 'testnet',
       reward
     );
+
+    if (!this._evaluationOptions.internalWrites && strict) {
+      const handlerResult = await this.callContract(input, interactionTx.owner, undefined, tags, transfer, strict, vrf);
+      if (handlerResult.type !== 'ok') {
+        throw Error(`Cannot create interaction: ${handlerResult.errorMessage}`);
+      }
+    }
+
     return interactionTx;
   }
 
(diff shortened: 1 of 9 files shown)


src/contract/HandlerBasedContract.ts
```diff
@@ -394,6 +387,14 @@ export class HandlerBasedContract<State> implements Contract<State> {
       this.warp.environment === 'testnet',
       reward
     );
+
+    if (!this._evaluationOptions.internalWrites && strict) {
```
koonopek, line 391: Move it here

src/contract/Signature.ts
```diff
@@ -2,34 +2,24 @@ import Transaction from 'arweave/node/lib/transaction';
 import { Warp } from '../core/Warp';
```
koonopek, line 2: Only refactor here, no logic changes

src/core/Warp.ts
```diff
@@ -90,7 +90,7 @@ export class Warp {
     return await this.createContract.register(id, bundlrNode);
   }
 
-  async createSourceTx(sourceData: SourceData, wallet: ArWallet | SignatureType): Promise<Transaction> {
+  async createSourceTx(sourceData: SourceData, wallet: ArWallet | CustomSignature): Promise<Transaction> {
```
koonopek, line 93: I have renamed `SignarureType` to `CustomSignature`

src/contract/HandlerBasedContract.ts
ppedziwiatr: hmm, at first glance - I guess I don't understand what issue this move is fixing?

koonopek: Before fix we were calling:
`const handlerResult = await this.callContract(input, undefined, undefined, tags, transfer, strict, vrf);`
notice second param (caller) is undefined, so then here:
https://github.com/warp-contracts/warp/blob/main/src/contract/HandlerBasedContract.ts#L581 we extracting owner by creating signature.

This move is fixing that we are calling now:
`const handlerResult = await this.callContract(input, interactionTx.owner, undefined, tags, transfer, strict, vrf);`
notice second param (caller) is extract from transaction so we don't enter this branch:
https://github.com/warp-contracts/warp/blob/main/src/contract/HandlerBasedContract.ts#L581

ppedziwiatr: oh, now I get it! thanks :-)

asiaziola: Hmm.. if the strict logic is moved after interaction creation can't we remove these lines:
```
      const handlerResult = await this.callContract(input, undefined, undefined, tags, transfer, strict, vrf);

      if (strict && handlerResult.type !== 'ok') {
        throw Error(`Cannot create interaction: ${handlerResult.errorMessage}`);
      }
```
in `internalWrites` condition? https://github.com/warp-contracts/warp/blob/main/src/contract/HandlerBasedContract.ts#L344

In this case `callContract` would be always called with owner defined and I guess we could remove then this dummy interaction. This would however required evaluating inner writes before strict mode can throw an error. I'm not an expert in internal writes feature, maybe it can have some performance issues.

asiaziola (approved):

PR #307, opened 2022-12-30
feat: custom plugin config

asiaziola:


src/core/WarpPlugin.ts
```diff
@@ -1,11 +1,10 @@
 export const knownWarpPlugins = [
   'evm-signature-verification',
-  'smartweave-extension-nlp',
-  'smartweave-extension-ethers',
   'subscription',
   'ivm-handler-api',
   'evaluation-progress',
-  'fetch-options'
+  'fetch-options',
+  `^smartweave-extension`
```
asiaziola, line 7: i have a hard time deciding what to do with this one. i see two options here:

1. leave it as is it is right now - as a regex in the `knownWarpPlugins` and test the plugin type against it

2. create a separate variable which will store `smartweave-extension-` string and check if plugin type starts with this string

3. my favourite way was to use typescript template literal type
```
type WarpExtensionPluginType = `smartweave-extension-${string}`
```
but I did not find a way to verify it in the runtime.

Don't know which option to choose -.-

Issue #290, opened 2022-12-06
fix: refactor the internal writes 


ppedziwiatr:
Introduce the "uncommited state" - each contract on the call tree for a given interaction should have its own uncommited state. This state should be used during the evaluation of the internal writes.
e.g. if Contract A writes to Contract B - and we're reading the state of the Contract B:
1. Read the state of the Contract B up to the internal write interaction
2. for the internal write interaction - set the uncommited state for contract B to current state
3. read the state of the contract A (the one that performs `write` on contract B)
4. at the time the Contract A tries to perform a write on Contract B - read Contract B state from the `contract.parent().uncommitedState`, where
contract = Contract A
contract.parent() = Contract B
 and perform the write.
The result of the write should be stored in `contract.parent().uncommitedState`
5. After returning with evaluiation to Contract B - load its state from contract.uncommitedState

Watch out while evaluating the writes from the 'caller' (i.e. the one that is performing a write) contract perspective.

At the end of the evaluation of each interaction of the 'root' contract (i.e. the that user is trying to read via `contract.readState()` - commit all "uncommited states" in the root contract's call tree (i.e. store the results in the SDK cache).

This change should also allow to remove the "inf. loop fix" (which protects agains falling into inf. call loop when contracts are interacting with each other (Contract A reads/writes to Contract B, where Contract B reads/writes to A, etc...) and simplify the evaluator code.

This should also allow to achieve https://github.com/warp-contracts/warp/issues/195

ppedziwiatr:
since 1.2.49

PR #311, opened 2023-01-04
Ppe/uncommitted state

ppedziwiatr:
https://github.com/warp-contracts/warp/issues/290 🤯


ppedziwiatr:
image
wtf

src/__tests__/integration/data/staking/erc-20.js
```diff
@@ -73,6 +73,10 @@ export function handle(state, action) {
     const recipient = _input.recipient;
     const amount = _input.amount;
 
+    if (amount == 0 ) {
+      throw new ContractError('TransferFromZero');
```
rpiszczatowski, line 77: Is amount the sender balance (transferFROMzero suggests that)? Shouldn't we check whether `amount <= 0`?

src/__tests__/integration/data/thethar/thethar-contract-wrc.js
```diff
@@ -0,0 +1,403 @@
...
+      return false;
+    };
+    var calcHash = (string) => {
+      var hash = 0, i, chr;
+      if (string.length === 0)
+        return hash;
+      for (i = 0; i < string.length; i++) {
+        chr = string.charCodeAt(i);
+        hash = (hash << 5) - hash + chr;
+        hash |= 0;
```
rpiszczatowski, line 29: Isn't `|= 0` basically noop?

src/__tests__/integration/data/thethar/thethar-contract-wrc.js
```diff
@@ -0,0 +1,403 @@
...
+      }
+      let sum = 0;
+      const r = await getRandomIntNumber(totalTokens);
+      for (const address of Object.keys(balances)) {
+        sum += balances[address];
+        if (r <= sum && balances[address] > 0) {
+          return address;
+        }
+      }
+      return void 0;
```
rpiszczatowski, line 46: What is the difference between `return void 0;` and `return;`?

ppedziwiatr: I believe that is a compiled version (i.e. version that was generated by some like esbuild or tsc compiler?).

I need to remove these files (an the rust junk), as they are not used effectively (Kuba was analysing the original issue on these files, but since then we've managed to isolate the original issue into a separate MRE).

src/__tests__/integration/data/thethar/thethar-contract-wrc.js
ppedziwiatr: again, it's a compiled code (probably generated by the TS compiler), so hard for me to say.

BUT...I believe the idea here is to use '0' as the 'default' value, in case the previous assignments would result in sth like NaN, null, or whatever.

src/__tests__/integration/data/staking/erc-20.js
ppedziwiatr: In a real contract - yes.

This contract is a simplified, stripped-down version (I believe there's a warn in the header) used only for testing certain features (like this specific case, where a transfer from zero was an indication of some issue in the internal writes feature in the whole test scenario)

src/__tests__/integration/internal-writes/internal-write-back.test.ts
```diff
@@ -279,7 +279,7 @@ describe('Testing internal writes', () => {
       expect((await contractB.readState()).cachedValue.state.counter).toEqual(2060);
     });
 
-    xit('should properly evaluate state with a new client', async () => {
+    it('should properly evaluate state with a new client', async () => {
```
ppedziwiatr, line 282: LFG!!!11111

PR #313, opened 2023-01-09
feat: allow closing cache connections

janekolszak:


src/core/Warp.ts
```diff
@@ -134,6 +134,15 @@ export class Warp {
     return this.plugins.get(type) as WarpPlugin<P, Q>;
   }
 
+  // Close cache connection
+  async close(): Promise<void> {
+    return Promise.all([
+      this.definitionLoader.getSrcCache().close(),
+      this.definitionLoader.getCache().close(),
+      this.stateEvaluator.getCache().close()
+    ]).then();
```
ppedziwiatr, line 143: y no `await Promise.all`?

ppedziwiatr (changes requested):

src/core/Warp.ts
ppedziwiatr: well, in fact even await is not needed, we can simply return a Promise...

janekolszak: That's what it does. It returns a Promise<void>. `.then()` is just to convert from `Promise<[void,void,void]> `

PR #325, opened 2023-01-17
fix: KV storage should use SortKeyCache, remove trie

Tadeuchi:


src/cache/SortKeyCache.ts
```diff
@@ -32,7 +32,14 @@ export interface SortKeyCache<V> {
   /**
    * removes contract's data
```
ppedziwiatr, line 33: update comment

src/cache/SortKeyCache.ts
```diff
@@ -22,7 +22,7 @@ export interface SortKeyCache<V> {
   /**
    * returns value for the key and exact blockHeight
```
ppedziwiatr, line 23: and exact sortKey - please update

src/cache/SortKeyCache.ts
```diff
@@ -10,9 +10,9 @@ export interface SortKeyCache<V> {
...
 
   /**
-   * returns latest value stored for given contractTxId
+   * returns latest value stored for a given key
    */
-  getLast(contractTxId: string): Promise<SortKeyCacheResult<V> | null>;
+  getLast(key: string): Promise<SortKeyCacheResult<V> | null>;
 
   /**
    * returns last cached sort key - takes all contracts into account
```
ppedziwiatr, line 18: in general - since you're fixing the naming - please update all the comments in this interface (e.g. "takes all contracts" here)

src/cache/impl/LevelDbCache.ts
```diff
@@ -94,21 +94,38 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+      if (op.type === 'put') {
+        await this.put(op.key, op.value);
+      } else if (op.type === 'del') {
+        await this.delete(op.key);
+      }
+    }
+  }
+
+  async open(): Promise<void> {
+    if (this.db.status === 'closed' || this.db.status === 'closing') {
```
ppedziwiatr, line 120: can you please a comment that using 'closing' here is safe - or maybe link the exact code from level library

src/cache/SortKeyCache.ts
```diff
@@ -22,7 +22,7 @@ export interface SortKeyCache<V> {
   /**
    * returns value for the key and exact blockHeight
    */
-  get(contractTxId: string, sortKey: string, returnDeepCopy?: boolean): Promise<SortKeyCacheResult<V> | null>;
+  get(key: string, sortKey: string): Promise<SortKeyCacheResult<V> | null>; //FIXME: why get by two values and put by object??
```
ppedziwiatr, line 25: please fix -> use the object

src/cache/impl/LevelDbCache.ts
```diff
@@ -94,21 +94,38 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+      await this.db.open();
+    }
+  }
+
+  async close(): Promise<void> {
+    if (this._db) {
```
ppedziwiatr, line 126: verify if some additional conditions on db.status are not required here

src/core/Warp.ts
```diff
@@ -67,8 +58,11 @@ export class Warp {
   ) {
     this.createContract = new DefaultCreateContract(arweave, this);
     this.testing = new Testing(arweave);
-    this.kvStorageFactory = (contractTxId: string) => {
-      return new TrieLevel(new Level(`${DEFAULT_LEVEL_DB_LOCATION}/kv/ldb/${contractTxId}`));
+    this.kvStorageFactory = (key: string) => {
```
ppedziwiatr, line 61: key -> contractTxId

src/core/Warp.ts
```diff
@@ -67,8 +58,11 @@ export class Warp {
...
+      return new LevelDbCache({
+        inMemory: false,
```
ppedziwiatr, line 63: set to true for warp.environement = 'local'

..and then remove manual cache deletion from the integration test.

src/legacy/smartweave-global.ts
```diff
@@ -243,31 +248,35 @@ class Vrf {
...
     this.checkStorageAvailable();
-    this._kvBatch.push({
-      type: 'del',
-      key: Buffer.from(key)
-    });
-  }
+    const sortKey = this._transaction.sortKey;
+
+    if (this._kvBatch.length > 0) {
+      const putBatches = this._kvBatch.filter((batchOp) => batchOp.type === 'put') as PutBatch<any>[];
```
ppedziwiatr, line 269: we should probably take 'del' operations into consideration here.
Ie. reverse and if the first operation for the given key is del - return null immediately.

ppedziwiatr (changes requested):

src/cache/impl/LevelDbCache.ts
```diff
@@ -94,21 +94,38 @@ export class LevelDbCache<V = any> implements SortKeyCache<V> {
...
+    const contractCache = this.db.sublevel<string, any>(key, { valueEncoding: 'json' });
     await contractCache.open();
     await contractCache.clear();
   }
 
-  close(): Promise<void> {
-    return null;
-    //return this.db.close();
+  async batch(opStack: BatchDBOp<V>[]) {
+    for (const op of opStack) {
```
ppedziwiatr, line 110: [db.batch](https://github.com/Level/level#dbbatchoperations-options-callback) ?

ppedziwiatr (changes requested):

src/cache/impl/LevelDbCache.ts
Tadeuchi: Yes, that's a good point, but there is a small issue with how we store values in sublevels.

For example we use transaction ids to get the sublevels and then sortKeys in sublevel.
If we would swich the transactionId with the sortKey the batch function would fit very well.

src/cache/impl/LevelDbCache.ts
Tadeuchi: It seems that leveldb is handling all these states pretty good when calling open/close and there seems to be no point in checking it.

src/cache/impl/LevelDbCache.ts
Tadeuchi: not required, calling close two times in a row is not causing any problems

src/core/Warp.ts
Tadeuchi: There are some side effects in this approach, some of the tests are not taking this change very well. Let's skip it for now.

src/legacy/smartweave-global.ts
Tadeuchi: We cannot just put nulls, and removing the whole key history is also not a option.
Let's consider adding some flag like: `removed` ?

src/cache/SortKeyCache.ts
Tadeuchi: Done, but I'm having second thoughts on this change...

ppedziwiatr: mind sharing? :>

src/cache/impl/LevelDbCache.ts
ppedziwiatr: yeah, fck it for now.

src/legacy/smartweave-global.ts
ppedziwiatr: del will be - maybe - added in a separate PR.

src/contract/HandlerBasedContract.ts
```diff
@@ -798,14 +799,19 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
-    const result: Map<string, string> = new Map();
+    const result: Map<string, StorageValue> = new Map();
     try {
       await storage.open();
       for (const key of keys) {
-        const value = await storage.get(Buffer.from(key));
-        result.set(key, value == null ? null : value.toString());
+        const lastValue = await storage.getLessOrEqual(key, lastSortKey);
+        const valStorage = {
+          keyStorage: lastSortKey,
```
ppedziwiatr, line 811: that's not very effective...ie. if we ask for 100 values, we will return the same value (the same sortKey) 100 times.

The returned object structure should be:
```
{ 
  sortKey,
  result
}
```

ppedziwiatr (changes requested):

src/contract/HandlerBasedContract.ts
Tadeuchi: yep, this is way better, fixed

Issue #264, opened 2022-11-19
feat: contract evaluation manifest


ppedziwiatr:
The contract deployment transaction should contain information about which options should be used to evaluate the contract (e.g, `allowBigInt`, `internalWrites`, etc) - stored in tags.

The SDK should automatically set all the required evaluation options based on this configuration.

The SDK should also WARN if the contract is not being evaluated with required options.

One potential issue here to solve is that each contract that takes part in the execution may require different evaluation options - what if - for example - parent contract 'blocks' the internal writes, but the child contract (e.g. the one whose parent contract is reading the state via `readContractState`) allows internal writes?

Currently - the whole evaluation inherits the evaluation options from the root contract settings.

ppedziwiatr:
The rules:
https://github.com/warp-contracts/warp/blob/ppe/grrrrrrrrrrrr/src/contract/EvaluationOptionsEvaluator.ts#L23

The tests showing how the most crucial options work:
https://github.com/warp-contracts/warp/blob/ppe/grrrrrrrrrrrr/src/__tests__/unit/evaluation-options.test.ts#L124

ppedziwiatr:
released in 1.2.35

jakub-wojciechowski:
Let's add new evaluation modes for:
1) Only Arweave L1
2) Only Sequencer/Bundles (L2)
3) Mix of both (L2 for speed + L1 for trustless exits)

jakub-wojciechowski:
It'll be great to consider signalling evaluation options also for existing contracts (maybe sth similar to the evolve mechanics).

ppedziwiatr:
What is the use case and which projects requested these features?

We spoke few months ago about restricting to only L1 or only sequencer (ar.io suggested such feature), but - afair - you were against it.
Also - I still don't see a real use case for point 3 (at least not in the Arweave ecosystem) - and it really adds a lot of complexity.

In general - this also adds some complexity to the interaction writing process (e.g. all interactions on mainnet warp instance - according to your request - are now sent by default to the sequencer. We would now need to first verify the contract manifest and choose the right method (not sure what to do in case of point 3)).
Additionally - the sequencer would now also need to verify if the contract supports it - i.e. it would need to first load contract data and verify its manifest - this will slow down the processing and sequencer's TPS.

btw., please don't reopen old, already implemented issues.
new features, new bugs, etc. = new issue.

ppedziwiatr:
> Let's add new evaluation modes for:
> 
[...]

PR #334, opened 2023-01-25
feat: contract evaluation manifest #264 - source type

Tadeuchi:


src/contract/HandlerBasedContract.ts
```diff
@@ -525,15 +524,8 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
-      this._eoEvaluator = new EvaluationOptionsEvaluator(
-        this.evaluationOptions(),
-        contractDefinition.manifest?.evaluationOptions
-      );
-    }
-    const contractEvaluationOptions = this.isRoot()
-      ? this._eoEvaluator.rootOptions
-      : this.getEoEvaluator().forForeignContract(contractDefinition.manifest?.evaluationOptions);
+
+    const contractEvaluationOptions = this.resolveEvaluationOptions(contractDefinition.manifest?.evaluationOptions);
```
ppedziwiatr, line 528: hmm, isn't it already done at line 502? That one might be optional, so I would move the contractEvaluationOptions before the big fuckin' if and check here if it hasn't be already set.

src/core/modules/StateEvaluator.ts
```diff
@@ -205,6 +208,8 @@ export interface EvaluationOptions {
   // an endpoint for retrieving wallet balance info
   walletBalanceUrl: string;
 
+  sourceType: SourceType;
```
ppedziwiatr, line 211: please add a comment what does this option represent.

src/__tests__/unit/evaluation-options.test.ts
```diff
@@ -1,5 +1,6 @@
 import { EvaluationOptionsEvaluator } from '../../contract/EvaluationOptionsEvaluator';
 import { WarpFactory } from '../../core/WarpFactory';
+import { SourceType } from '../../core/modules/impl/WarpGatewayInteractionsLoader';
```
ppedziwiatr, line 3: I would add some dumb intergration test, that would deploy a contract with a manifest set to sourceType: sequencer, and then create some transactions (which in case of local testing are always considered as 'direct' Arweave transactions) - and verify that the SDK always does not load anything.

src/contract/HandlerBasedContract.ts
Tadeuchi: sure

src/core/modules/StateEvaluator.ts
Tadeuchi: done

src/__tests__/unit/evaluation-options.test.ts
Tadeuchi: As discussed this is not that easy. It will probably require dedicated config for local env and some fetch mocking. And after all that it may not be that reliable.

ppedziwiatr (approved):

PR #337, opened 2023-01-25
feat: expose KV to WASM

rpiszczatowski:


src/core/modules/impl/wasm/rust-wasm-imports.ts
```diff
@@ -268,6 +291,14 @@ export const rustWasmImports = (swGlobal, wbindgenImports, wasmInstance, dtorVal
       var ret = getStringFromWasm0(arg0, arg1);
       return addHeapObject(ret);
     },
+    __wbindgen_string_get: function(arg0, arg1) {
```
ppedziwiatr, line 294: hmm, weird that this function was not needed before?

Why is it needed now?

rpiszczatowski: `__wbindgen_XXX` methods result from calling methods on `JsValue` instances. This particular one comes from `JsValue::as_string`. I think that we should probably add all those methods to our glue code as today some of `JsValue` methods will simply panic.

ppedziwiatr: yes, please.

Issue #248, opened 2022-11-04
fix: extract VRF to a plugin


ppedziwiatr:
..in order to remove crypto deps from the core sdk and reduce bundle size.

ppedziwiatr:
fixed within https://github.com/warp-contracts/warp/pull/401

PR #347, opened 2023-02-05
fix: fixing deps to arweave-js

ppedziwiatr:
1. ~moving arweave-js as peer dependency (not sure about this)~
2. fixing warp-wasm-json-toolkit (was using imports from redstone-isomorphic)
3. fixing imports to bn.js (multiple packages were importing its own version 🤦 )
4. reducing imports from arweave-js (and fixing issue with adding both node and web version in the resulting bundle 🤦 )
5. fixed imports to bignumber.js (multiple packages were importing its own version 🤦 )
6. esbuild version update
7. fixed imports to buffer-pipe (multiple packages were importing its own version 🤦 )

After (before it was around 750 kb 😮 ):
image

The biggest part of the node_modules is related to VRF - it needs to be moved to a plugin - https://github.com/warp-contracts/warp/issues/248


ppedziwiatr:
`1.2.52-beta.2`

PR #355, opened 2023-02-21
fix: caching state - store latest value

ppedziwiatr:


src/core/modules/impl/DefaultStateEvaluator.ts
```diff
@@ -287,9 +287,13 @@ export abstract class DefaultStateEvaluator implements StateEvaluator {
         }
       }
 
+      // if that's the end of the root contract's interaction - commit all the uncommitted states to cache.
       if (contract.isRoot()) {
+        // update the uncommitted state of the root contract
+        contract.setUncommittedState(contract.txId(), lastConfirmedTxState.state);
```
Tadeuchi, line 293: IDE is suggesting that the lastConfirmedTxState could be null, and considering the line 300 with null checking this suggestion seems to be reasonable.
`if (lastConfirmedTxState !== null) {`

src/__tests__/regression/test-cases/read-state.json
```diff
@@ -95,13 +94,9 @@
   "8zGVX17V6u3Uzn2fWJFya3vEuPOvIZ6FRifUDhaprtE",
   "Zwp7r7Z10O0TuF6lmFApB7m5lJIrE5RbLAVWg_WKNcU",
   "1enYq8LDJQ7fxW20WMWxbq3jpMaBeN-k67WXnhEDIds",
-  "efrXSmKuDH32Svllu9g15PxTlxrL5T3UNT1jxgQqat0",
```
ppedziwiatr, line 98: all those contracts were using unsafeClient to load tx data from arweave and were causing random test errors.

PR #362, opened 2023-02-28
feat: add constructor for js contracts

koonopek:


src/__tests__/integration/basic/constructor.test.ts
```diff
@@ -0,0 +1,225 @@
...
+            expect(state.counter).toBe(2);
+
+            await contract.writeInteraction({ function: 'nop' })
+            await contract.writeInteraction({ function: 'nop' })
+
+            const { cachedValue: { state: state2 } } = await contract.readState();
+            expect(state2.counter).toStrictEqual(2);
+        });
+
+        describe.skip('TODO: doesn"t work internal writes', () => {
```
koonopek, line 159: Here is my biggest nightmare, internal writes doesn't work.
I'm not sure why, maybe because `contractDefinition.contractTx` doesn't have assigned `sortKey` and internalWrites doesn't know if it should evaluate it

src/core/modules/impl/handler/WasmHandlerApi.ts
```diff
@@ -125,6 +125,13 @@ export class WasmHandlerApi<State> extends AbstractContractHandler<State> {
     }
   }
 
+  maybeCallStateConstructor(initialState: State, executionContext: ExecutionContext<State, unknown>): Promise<State> {
+    if (this.contractDefinition.manifest?.evaluationOptions.useConstructor) {
+      throw Error("Constructor is not implemented for wasm");
+    }
+    return Promise.resolve(initialState);
+  }
```
koonopek, line 133: Not implemented yet for rust.

src/legacy/smartweave-global.ts
```diff
@@ -142,7 +142,7 @@ export class SmartWeaveGlobal {
 
 // tslint:disable-next-line: max-classes-per-file
```
koonopek, line 143: formatting only

src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -75,18 +110,35 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
...
   initState(state: State): void {
-    // nth to do in this impl...
+
+  }
+
+  async maybeCallStateConstructor<Input>(initialState: State, executionContext: ExecutionContext<State>): Promise<State> {
+    if (this.contractDefinition.manifest?.evaluationOptions.useConstructor) {
+      const interaction = { input: { function: INIT_FUNC_NAME, args: initialState } as Input, caller: this.contractDefinition.owner };
+      // TODO: we have to use some default sort key like 0,0,0,0
+      const interactionData: InteractionData<Input> = { interaction, interactionTx: { ...this.contractDefinition.contractTx, sortKey: '0.0.0.0' } };
```
koonopek, line 125: Hack with `sortKey` to make KV works, however it need fix now.
I see two options:
1) don't enable KV and internal writes in constructor (bad user experience)
2) generate some hacky sortKey which will be always first 
3) make constructor sperate transaction (need additional signing)
4) dont' know

koonopek: Hmmmmmm, internal writes are impossible to implement in constructor if we don't produce new interaction (I think we don't wont to do this).
So new options:
Change implementation which produce new interaction `__init` (a little bit shady, more code)
Leave as is, with working KV in `__init` function (constructor), but without support for KV (only in __init function)

src/__tests__/integration/basic/constructor.test.ts
koonopek: Internasl writes won't be working in constructor

src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -54,39 +61,111 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
...
+  async maybeCallStateConstructor<Input>(
+    initialState: State,
+    executionContext: ExecutionContext<State>
+  ): Promise<State> {
+    if (this.contractDefinition.manifest?.evaluationOptions.useConstructor) {
+      const interaction = {
+        input: { function: INIT_FUNC_NAME, args: initialState } as Input,
+        caller: this.contractDefinition.owner
+      };
+      const interactionTx = { ...this.contractDefinition.contractTx, sortKey: ZERO_SORT_KEY };
```
koonopek, line 155: Zero sort key hack is for enabling KV to work in __init

koonopek:
Integration tests not passing cause of 404 :(
passing locally

src/core/modules/StateEvaluator.ts
```diff
@@ -236,6 +238,8 @@ export interface EvaluationOptions {
   // whether a separate key-value storage should be used for the contract
   useKVStorage: boolean;
 
+  useConstructor: boolean;
```
ppedziwiatr, line 241: can you please add a comment above the new parameter? explaining what exactly given parameter means.

src/__tests__/integration/basic/constructor.test.ts
```diff
@@ -0,0 +1,245 @@
...
+                        },
+                        "type": "ok",
+                    }
+                )
+
+            });
+        });
+    });
+
+    describe('with useConstructor = false', () => {
```
ppedziwiatr, line 202: please fix formatting in tests (4tabs vs 2tabs, etc).

src/contract/EvaluationOptionsEvaluator.ts
```diff
@@ -104,9 +104,10 @@ export class EvaluationOptionsEvaluator {
     walletBalanceUrl: () => this.rootOptions['walletBalanceUrl'],
     mineArLocalBlocks: () => this.rootOptions['mineArLocalBlocks'],
     cacheEveryNInteractions: () => this.rootOptions['cacheEveryNInteractions'],
+    useKVStorage: (foreignOptions) => foreignOptions['useKVStorage'],
```
ppedziwiatr, line 107: I believe the new parameter should be also added in if in line 122 - otherwise it will probably not work in DRE.

src/core/modules/impl/CacheableStateEvaluator.ts
```diff
@@ -49,28 +49,34 @@ export class CacheableStateEvaluator extends DefaultStateEvaluator {
...
+        executionContext.contractDefinition.initState,
+        executionContext
+      );
+    }
+
     if (missingInteractions.length == 0) {
       this.cLogger.info(`No missing interactions ${contractTxId}`);
       if (cachedState) {
-        executionContext.handler?.initState(cachedState.cachedValue.state);
+        executionContext.handler.initState(cachedState.cachedValue.state);
```
ppedziwiatr, line 67: hmm, with your refactor above - I believe it could be now called like
`executionContext.handler.initState(baseState);` before line 66?

ppedziwiatr: or maybe even MUST be called like this, otherwise you might loose changes set in the constructor (line 58)

ppedziwiatr: also - why `?` has been removed? handler is not always created and added to executionContext.

src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -1,11 +1,24 @@
...
+const throwErrorWithName = (name: string, message: string) => {
+  const error = new Error(message);
+  error.name = name;
+  throw error;
+};
+enum KnownErrors {
+  ContractError = 'ContractError',
+  ConstructorError = 'ConstructorError'
+}
+const ZERO_SORT_KEY = '000000000000,0000000000000,0000000000000000000000000000000000000000000000000000000000000000';
```
ppedziwiatr, line 20: I believe we already have a const for this - https://github.com/warp-contracts/warp/blob/main/src/core/modules/impl/LexicographicalInteractionsSorter.ts#L16

src/core/modules/impl/handler/JsHandlerApi.ts
ppedziwiatr: so all the values set in KV in the constructor will have the 'zero' sortKey set, right?

src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -16,36 +29,27 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
     super(swGlobal, contractDefinition);
   }
 
-  async handle<Input, Result>(
+  private async runContractFunction<Input>(
```
ppedziwiatr, line 32: can you please move private functions to the bottom of the class (i.e. first public, than private)

src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -54,39 +58,113 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
...
       switch (err.name) {
-        case 'ContractError':
+        case KnownErrors.ContractError:
           return {
-            type: 'error',
+            type: 'error' as const,
             errorMessage: err.message,
-            state: currentResult.state,
+            state: state,
             // note: previous version was writing error message to a "result" field,
```
ppedziwiatr, line 66: since you're cleaning-up this class - please remove this comment as well.

src/legacy/smartweave-global.ts
```diff
@@ -257,7 +257,10 @@ class Vrf {
 export class KV {
   private _kvBatch: BatchDBOp<any>[] = [];
 
-  constructor(private readonly _storage: SortKeyCache<any> | null, private readonly _transaction: Transaction) {}
+  constructor(
```
ppedziwiatr, line 260: hmm, is this our standard linter formatting?

src/__tests__/integration/data/constructor/constructor.js
```diff
@@ -0,0 +1,18 @@
...
+        await SmartWeave.kv.put("__init", SmartWeave.transaction.id);
+
+        if (action.input.args.fail) {
+            throw new ContractError("Fail on purpose")
+        }
+        state.counter = action.input.args.counter + 1;
+    }
+
+    return { state }
+}
```
ppedziwiatr, line 18: new lines...

ppedziwiatr: I believe VSCode has a plugin for this :-)

src/core/modules/impl/handler/JsHandlerApi.ts
koonopek: Right

src/core/modules/StateEvaluator.ts
koonopek: Sure I have done this, someone must have eaten it :)
Added :)

src/contract/EvaluationOptionsEvaluator.ts
koonopek: Truue, nice catch, thanks

src/core/modules/impl/CacheableStateEvaluator.ts
koonopek: true true, would fail for rust contracts, which are not implemented for now.
I have removed `?` because `ExecutionContext::handler` is not defined as optional: `handler: Api`

src/core/modules/impl/handler/JsHandlerApi.ts
koonopek: fixed

src/core/modules/impl/handler/JsHandlerApi.ts
koonopek: of course

src/legacy/smartweave-global.ts
koonopek: Nope, I done my self, because parameters were not fitting into my wide 13inch screen
However I have ran linter - ohhh i see we have also prettier :) Now everything should be properly formatted :D

src/__tests__/integration/data/constructor/constructor.js
koonopek: Dicussed above

src/core/modules/impl/CacheableStateEvaluator.ts
koonopek: Nope it shouldn't called like this because it means if(cachedState) it means that this is not first interaction so baseState === null (constructor won't be called)

koonopek:
Everything is great again: 🇺🇸 
image

src/core/modules/impl/CacheableStateEvaluator.ts
ppedziwiatr: ok, mea culpa!

koonopek:
@ppedziwiatr  docs for constructor, I can merege after your approve https://github.com/warp-contracts/academy/pull/66

src/__tests__/integration/basic/constructor.test.ts
```diff
@@ -0,0 +1,267 @@
...
+
+        expect(state2.calls).toBeDefined();
+        expect(state2.calls).toEqual(['nop', 'nop', 'nop']);
+      });
+    });
+
+    it('should NOT fail to call __init function', async () => {
+      const contract = await deployContract({ withConstructor: false });
+
+      await expect(contract.writeInteraction({ function: '__init' })).resolves.toBeDefined();
```
koonopek, line 264: Here was missing `toBeDefined()` and that is way tests were failing, something specific to jest matchers :)

ppedziwiatr (approved):

Issue #345, opened 2023-02-03
feat: Add functions to the Kv API in order to query ranges of keys and/or values 


noomly:
Exposing the [keys](https://github.com/Level/level#keyiterator--dbkeysoptions) and the [values](https://github.com/Level/level#valueiterator--dbvaluesoptions) functions that leveldb provide would already be very useful in `Kv`. I imagine it'd not be too hard to write an equivalent feature for other key-value stores (like lmdb for example) that could fit into the same API.

Exposing the iterator api would be extremely useful for kv databases that are organized in a hierarchical way. Take this key structure for example:

```
settings.contractName
settings.operators
tokens.{}.ticker
tokens.{}.balances.{}
```

`{}` being any kind of string, here ids or addresses. In this example, it might be useful to get every balances for a specific token in order to calculate its current supply. This would be achieved pretty easily with the `values` function, like so:

```typescript
const tokenPath = `tokens.aRandomTokenId.balances.`;
const supply = (
    await db
        .values({
            gte: tokenPath,
            lte: `${tokenPath}\xff`,
        })
        .all()
).reduce((acc, value) => acc + parseInt(value), 0);
```

ppedziwiatr:
I guess the lmdb-js has similar APIs (https://github.com/kriszyp/lmdb-js#dbgetvalueskey-options-rangeoptions-iterableany , https://github.com/kriszyp/lmdb-js#dbgetkeysoptions-rangeoptions-iterableany ) so it should be doable...

@Tadeuchi

Tadeuchi:
Hey @noomly, I'm trying to implement the range queries for kv storage. 
I created some draft and I was wondering if you could take a look and let me know if this is something that you had in mind. 

You can find my PR above, and also some simple interaction example that uses the range options in kvstorage 
https://github.com/warp-contracts/warp/pull/364/files#diff-b89ef4e5c5624ccac706296f669964e1b2f31b2a66dd49537ebffcd11780610f

noomly:
Hey @Tadeuchi, the `keys` and `entries` method are exactly like what I had in mind, they should be enough to get started. Thank you!

I'm just a bit curious about the usage of the `dbSeparator` prepended with the `sortKey`, are different versions of the db stored inside a single KV for a contract?

Tadeuchi:
Unfortunately including keys and values from undergoing interaction in range queries is not that simple… 
I’ll explain the whole issue.

[...]

PR #364, opened 2023-03-03
feat: Add functions to the Kv API in order to query ranges of keys an…

Tadeuchi:
…d/or values #345

### DIFF
--- a/src/__tests__/integration/data/kv-storage-range.js
+++ b/src/__tests__/integration/data/kv-storage-range.js
@@ -1,0 +1,74 @@
+export async function handle(state, action) {
+  const input = action.input;
+  const caller = action.caller;
+
+  if (input.function === 'mint') {
+    await SmartWeave.kv.put('mint.' + input.target, input.qty);
+    await SmartWeave.kv.put(input.target, input.qty);
+
+    return {state};
+  }
+
+  if (input.function === 'transfer') {
+    const target = input.target;
+    const qty = input.qty;
+
+    if (!Number.isInteger(qty)) {
+      throw new ContractError('Invalid value for "qty". Must be an integer');
+    }
+
+    if (!target) {
+      throw new ContractError('No target specified');
+    }
+
+    if (qty <= 0 || caller === target) {
+      throw new ContractError('Invalid token transfer');
+    }
+
+    let callerBalance = await SmartWeave.kv.get(caller);
+    callerBalance = callerBalance ? callerBalance : 0;
+
+    if (callerBalance < qty) {
+      throw new ContractError(`Caller balance not high enough to send ${qty} token(s)!`);
+    }
+
+    // Lower the token balance of the caller
+    callerBalance -= qty;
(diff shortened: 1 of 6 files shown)


src/legacy/smartweave-global.ts
```diff
@@ -286,6 +287,16 @@ export class KV {
     return result?.cachedValue || null;
   }
 
+  async keys(options?: SortKeyCacheRangeOptions): Promise<string[]> {
+    const sortKey = this._transaction.sortKey;
+    return this._storage.keys(sortKey, options)
```
Tadeuchi, line 292: should also include keys from _kvBatch

src/cache/impl/LevelDbCache.ts
```diff
@@ -154,14 +157,48 @@ export class LevelDbCache<V> implements SortKeyCache<V> {
...
+      .filter((k) => !sortKey || this.extractSortKey(k).localeCompare(sortKey) <= 0)
+      .map((k) => this.extractOriginalKey(k))
+      .forEach((k) => distinctKeys.add(k));
+
+    return Array.from(distinctKeys);
+  }
 
-    return Array.from(result);
+  extractOriginalKey(joinedKey: string): string {
+    return joinedKey.split('!')[1];
```
Tadeuchi, line 173: maybe first char in joined key can be treated as a separator

src/__tests__/integration/data/kv-storage-range.js
```diff
@@ -0,0 +1,74 @@
...
+    return {result: {target, ticker, balance: result ? result : 0}};
+  }
+
+  if (input.function === 'minted') {
+
+    for (const entry of await await SmartWeave.kv.entries()) {
+      console.log(`getting value ${entry.value} for key ${entry.key}`)
+    }
+
+    const sumMinted = (await SmartWeave.kv.entries({ gte: 'mint.', lte: 'mint.\xff'}))
```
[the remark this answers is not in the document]
ppedziwiatr: on 2nd. thought - 'optimistic' writes and removing values on rollbacks is probably a bad solution, e.g.:

1. A.x makes a 'write' on B
2. B.x (i.e. the 'write' from point 1.) makes a 'put' in KV Storage
3. A.x makes another write on B
3. B.x. (i.e. the write from point 3) makes another 'put' in KV Storage (with the same 'key' as in point 2).
4. Now B.x fails
5. A.x makes a viewState on B, which checks the value of the key set in points 2. and 3.

If we would decide to remove the values, the rollback from point 4. would cause the values for the key and for the 'x' sortKey to be removed completelty - i.e. the viewState would remove null - where it _should_ return value that was set in point 2.

ppedziwiatr:
Another issue with the 'range' feature - if we decide to store information about deleted values, we will need to take it into account while calculating the values for the given range...wonder how big of an issues this is...I guess in a simplest form we would need to simply manually remove the values from what's the current implementation returns...

PR #368, opened 2023-03-10
fix: make arweave interaction loader great again 🗽

koonopek:


src/__tests__/integration/basic/arweave-gateway-interaction-loader.test.ts
```diff
@@ -0,0 +1,71 @@
+import { writeFileSync } from 'fs';
```
koonopek, line 1: This will be removed form integrations tests, just to show you @ppedziwiatr  example that it works

koonopek:
To sum up:
- I didn't manage to make arweaveContractDefinitonLoader works, it can't find tx (but can find same tx Data..)
- Sorting is deterministic, however results doesn't have to be, because txData differs (timestamp, blockHeight) etc. We would have to provide them in tags to make it work
- used contract is our implementation of atomic-asset in typescript

koonopek:
Left:
1. `Warp.useCustom` 
2. Example in academy
3. refactor tags enum

I have also disabled access to transaction from constructor in this branch. Because I realised that it is not perfectly standardised and arlocal and gateway return diffrent transactions, which may lead to different results in test environment(local) vs mainnet.

src/core/modules/impl/ArweaveGQLTxsFetcher.ts
```diff
@@ -0,0 +1,138 @@
...
+  after?: string;
+}
+
+const RETRY_TIMEOUT = 30 * 1000;
+const MAX_REQUEST = 100;
+
+/**
+ * Query all transactions from arweave gateway
+ */
+export class ArweaveGQLTxsFetcher {
```
koonopek, line 82: Helper for quering gql for transactions

src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -120,8 +120,14 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
...
     });
 
+    const disabledTransaction = new Proxy(this.swGlobal.transaction, {
+      get: () =>
+        throwErrorWithName('ConstructorError', 'SmartWeave.transaction object is not accessible in constructor')
+    });
+
     this.swGlobal.vrf = disabledVrf;
     this.swGlobal.block = disabledBlock;
+    this.swGlobal.transaction = disabledTransaction;
```
koonopek, line 130: Disabled read comment below

ppedziwiatr (approved):

src/core/modules/impl/handler/JsHandlerApi.ts
koonopek: Left:

Warp.useCustom
Example in academy
refactor tags enum
I have also disabled access to transaction from constructor in this branch. Because I realised that it is not perfectly standardised and arlocal and gateway return diffrent transactions, which may lead to different results in test environment(local) vs mainnet.

koonopek:
DONT MERGE

PR #369, opened 2023-03-14
Robal/gh 309 warp contract macro

rpiszczatowski:


rust-warp-contracts/src/foreign_call.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub async fn write_foreign_contract<I: Serialize, E: DeserializeOwned + Debug>(
+    contract_address: &str,
+    input: I,
+) -> WriteResult<(), E> {
+    let input = match to_json_value(&input) {
+        Ok(v) => v,
+        Err(e) => return WriteResult::RuntimeError(format!("{e:?}")),
+    };
+    let write_result = SmartWeave::write(contract_address, input).await;
```
ppedziwiatr, line 55: verify if the state changes made via write are immediately available (e.g. when checked via viewState)

Issue #262, opened 2022-11-19
feat: throw if bundled tx tries to acces tx.reward or tx.transfer data


ppedziwiatr:
The contract interactions processed via bundled transactions MUST NOT access any data releated to transactions reward and transactions transfers (as in this case this data is irrelevant).

To secure against dumb progamming errors in the contracts error, the SmartWeaveGlobal object should throw a ContractError whenever a contract code tries to access:
`SmartWeave.transaction.target`
`SmartWeave.transaction.quantity`
`SmartWeave.transaction.reward`

The global `SmartWeave` object should probably also contain information whether given interaction comes from a bundled transaction (a boolean in `SmartWeave.transaction.bundled` ?)

PR #370, opened 2023-03-14
fix: smartweave global in constructor

koonopek:
So I have notice (a little bit too late), that `contractTx` differs from `interactionTx`.
Diffs:
1. missing sequencer tags wich can be used to create SmartWeave.block data (timestamp,height,id)
2. missing `block` data , when fetched from warp-gateway or arweave-gateway
3. also `vrf`object won't be accessible for now, it can be in future if required

- 1. and 2. can be fixed by adding extra tags to deployment tx by sequencer, (when sequencer is not used extra query will be needed)
- 3. Is also fixable, however we don't know if anybody will use it. So we can wait for community requesting this feature.

This MR:
- I disabled vrf and block access by using proxy to provide understable error, I will add info in academy (edit added)

### DIFF
--- a/src/core/modules/impl/handler/JsHandlerApi.ts
+++ b/src/core/modules/impl/handler/JsHandlerApi.ts
@@ -51,17 +51,20 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
     initialState: State,
     executionContext: ExecutionContext<State>
   ): Promise<State> {
-    if (this.contractDefinition.manifest?.evaluationOptions.useConstructor) {
+    if (this.contractDefinition.manifest?.evaluationOptions?.useConstructor) {
       const interaction = {
         input: { function: INIT_FUNC_NAME, args: initialState } as Input,
         caller: this.contractDefinition.owner
       };
-      const interactionTx = { ...this.contractDefinition.contractTx, sortKey: genesisSortKey };
-      // this is hard corded sortKey to make KV possible
+
+      const interactionTx = (await this.makeInteractionTxFromContractTx(
+        this.contractDefinition.contractTx,
+        this.contractDefinition.owner
+      )) as GQLNodeInterface;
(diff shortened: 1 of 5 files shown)


src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -73,16 +76,33 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
     }
   }
 
+  // interaction Tx is not
```
ppedziwiatr, line 79: I sense much fear in you...

src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -92,6 +112,17 @@ export class JsHandlerApi<State> extends AbstractContractHandler<State> {
       throwErrorWithName('ConstructorError', templateErrorMessage('refreshState'));
     this.swGlobal.contracts.viewContractState = () =>
       throwErrorWithName('ConstructorError', templateErrorMessage('viewContractState'));
+
+    const disabledVrf = new Proxy(this.swGlobal.vrf, {
```
ppedziwiatr, line 116: I like the Proxy idea, we could probably use it also for implementing https://github.com/warp-contracts/warp/issues/262

src/core/modules/impl/handler/JsHandlerApi.ts
koonopek: :))))))

src/core/modules/impl/handler/JsHandlerApi.ts
koonopek: Yep, however i would consider implementing different smartWeaveGlobal for different contexts, to be more declarative it could be hard to trace proxies in code.

ppedziwiatr (approved):

Issue #220, opened 2022-09-08
Contract deployment module should use the warp gw address passed in the configuration


ppedziwiatr:
...currently it is using hardcoded value from the WARP_GW_URL constant.

ppedziwiatr:
this crap is so fuckin' annoying, @Tadeuchi you have to fix this!!!!

The same value should be used
1. during contract deployment
2. during interactions load
3. during new interaction deployment
4. and probably one more place, which I can't recall right now :-)

ppedziwiatr:
fixed within https://github.com/warp-contracts/warp/commit/5ffc438749577263e8586a263ee97f02606f8ca4

PR #386, opened 2023-04-04
feat: Contract deployment module should use the warp gw address passe…

ppedziwiatr:
…d in the configuration #220


src/utils/utils.ts
```diff
@@ -86,3 +86,20 @@ export function bufToBn(buf: Buffer) {
...
+export async function getJsonResponse<T>(response: Promise<Response>): Promise<T> {
+  let r;
+  try {
+    r = await response;
+  } catch (e) {
+    throw new Error(`Error while communicating with gateway: ${JSON.stringify(e)}`);
+  }
+
+  if (!r?.ok) {
+    const text = await r.text();
```
asiaziola, line 99: works quite ok for errors thrown through the koa context like this one:
`ctx.throw(400, 'Error message');`

the error thrown by sdk:
`Error: 400: {"message":{"message":"Contract data item binary is not valid."}}`

for the errors thrown through regular Error object:
`throw new Error('Error text.')`

the error thrown by sdk:
`Error: 500: {"message":{}}`

I'm not sure if we could always pass error message in the ctx.body when error is thrown? like this:
`ctx.body = e.message`

src/core/modules/impl/WarpGatewayContractDefinitionLoader.ts
```diff
@@ -67,20 +65,11 @@ export class WarpGatewayContractDefinitionLoader implements DefinitionLoader {
...
-        .catch((error) => {
-          if (error.body?.message) {
-            this.rLogger.error(error.body.message);
-          }
-          throw new Error(
-            `Unable to retrieve contract data. Warp gateway responded with status ${error.status}:${error.body?.message}`
-          );
-        });
+      const baseUrl = stripTrailingSlash(this._warp.gwUrl());
+      const result: ContractDefinition<State> = await getJsonResponse(
```
asiaziola, line 69: await getJsonResponse<ContractDefinition<State>>

src/utils/utils.ts
ppedziwiatr: the last case is be fixed by the changes in the GW - https://github.com/warp-contracts/gateway/pull/161

src/core/modules/impl/WarpGatewayContractDefinitionLoader.ts
ppedziwiatr: nope.

PR #388, opened 2023-04-05
Evolve state in rust contracts

rpiszczatowski:
Rust implementation of evolve_state concept

### DIFF
--- a/crates/pst/contract/definition/src/state.rs
+++ b/crates/pst/contract/definition/src/state.rs
@@ -12,8 +12,7 @@ pub struct PstState {
     pub owner: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub evolve: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub can_evolve: Option<bool>,
+    pub can_evolve: bool,
     pub balances: HashMap<String, u64>,
 }
 
--- a/crates/warp-contracts/warp-contracts-core/src/evolve_state.rs
+++ b/crates/warp-contracts/warp-contracts-core/src/evolve_state.rs
@@ -1,0 +1,13 @@
+use serde::Serialize;
+
+#[derive(Serialize, Debug)]
+pub struct EvolveState {
+    pub can_evolve: bool,
+    pub source_transaction_id: Option<String>,
+}
+
+impl Default for EvolveState {
+    fn default() -> Self {
+        EvolveState { can_evolve: true, source_transaction_id: None }
+    }
+}
--- a/src/core/modules/impl/handler/WasmHandlerApi.ts
+++ b/src/core/modules/impl/handler/WasmHandlerApi.ts
@@ -90,6 +90,14 @@ export class WasmHandlerApi<State> extends AbstractContractHandler<State> {
     }
   }
 
+  evolveState() {
+    if ('evolveState' in this.wasmExports) {
+      return this.wasmExports.evolveState();
+    } else {
+      return this.doGetCurrentState();
+    }
+  }
+
   private async doHandleLegacy(action: ContractInteraction<unknown>): Promise<any> {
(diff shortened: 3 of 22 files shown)


src/core/modules/impl/HandlerExecutorFactory.ts
```diff
@@ -266,6 +267,8 @@ export interface HandlerApi<State> {
   initState(state: State): void;
 
   maybeCallStateConstructor(initialState: State, executionContext: ExecutionContext<State>): Promise<State>;
+
+  evolveState(): Promise<EvolveState>;
```
ppedziwiatr, line 271: Why is it here? Shouldn't it rather be in HandlerApi interface?

crates/warp-contracts/warp-contracts-core/src/evolve_state.rs
```diff
@@ -0,0 +1,17 @@
+use serde::Serialize;
+
+#[derive(Serialize, Debug)]
+pub struct EvolveState {
+    pub can_evolve: bool,
```
ppedziwiatr, line 5: will this map properly to fields in json state, i.e. 'canEvolve' and 'evolve'?

src/core/modules/impl/handler/WasmHandlerApi.ts
```diff
@@ -90,6 +90,14 @@ export class WasmHandlerApi<State> extends AbstractContractHandler<State> {
     }
   }
 
+  evolveState() {
```
ppedziwiatr, line 93: is it possible to map the return type? at least sth like `EvolveState | State`?

src/core/modules/impl/HandlerExecutorFactory.ts
rpiszczatowski: wow. This shouldn't be here. I'll remove it.

src/core/modules/impl/handler/WasmHandlerApi.ts
rpiszczatowski: I don't understand the question. I added this method only to demonstrate how to use new API. Feel free to change it in whatever way you want.

crates/warp-contracts/warp-contracts-core/src/evolve_state.rs
rpiszczatowski: No, it won't. This is supposed to be fetched by SDK WasmHandlerApi and process there. I didn't see reasons to make these field names compatible with our usual State fields. But of course it can be done.

crates/pst/contract/definition/src/state.rs
```diff
@@ -17,6 +18,15 @@ pub struct PstState {
     pub balances: HashMap<String, u64>,
 }
 
+impl Evolve for PstState {
```
ppedziwiatr, line 21: How will the contract authors know that they need to implement this trait?

I thought about sth like:

```
pub struct PstState {
    #[serde(flatten)]
    pub evolve: EvolveState,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub name: Option<String>,
    pub owner: String,
    pub balances: HashMap<String, u64>,
}
```

- the evolve would be a struct that could be composed with contract-specific state...

or maybe a macro?
```
#[warp-contract(evolve)]
pub struct PstState {
    #[serde(skip_serializing_if = "Option::is_none")]
    pub name: Option<String>,
    pub owner: String,
    pub balances: HashMap<String, u64>,
}
```

- that automagically adds required fields to the State struct and also implements the `Evolve` trait (which I believe will always look the same, ie. will have the same mapping:
```
can_evolve: self.can_evolve.unwrap_or(true),
source_transaction_id: self.evolve.clone(),
```
)

rpiszczatowski: compilation will fail saying that required trait is not implemented.

PR #399, opened 2023-04-21
adding moduleNameMapper to jest config so that warp workdir is includ…

Tadeuchi:
…ed in modules resolve


package.json
```diff
@@ -35,28 +33,27 @@
...
-    "test:unit:cache": "file=./src/__tests__/unit/cache-leveldb.test.ts yarn test",
-    "test:unit:cache:real": "file=./src/__tests__/unit/cache-leveldb-real-data.test.ts yarn test",
-    "test:integration": "file=./src/__tests__/integration yarn test",
-    "test:integration:basic": "file=./src/__tests__/integration/basic yarn test",
-    "test:integration:basic:load": "--silent=false --detectOpenHandles file=./src/__tests__/integration/basic/contract-loading.test.ts yarn test",
-    "test:integration:basic:arweave": "file=./src/__tests__/integration/basic/arweave-transactions-loading yarn test",
-    "test:integration:internal-writes": "file=./src/__tests__/integration/internal-writes yarn test",
-    "test:integration:wasm": "file=./src/__tests__/integration/wasm yarn test",
-    "test:regression": "yarn install:warp && node ./node_modules/.bin/jest ./src/__tests__/regression && yarn remove:warp",
+    "test": "jest",
```
ppedziwiatr, line 36: hmm, makes no sense to me now :-) I guess it should be removed as well, since all the 'test...' runs use the jest directly.

jest.config.js
```diff
@@ -4,18 +4,13 @@ module.exports = {
...
+  testPathIgnorePatterns: ['/.yalc/', '/data/', '/_helpers'],
 
   testEnvironment: 'node',
 
-  "transformIgnorePatterns": [
-    "<rootDir>/node_modules/(?!@assemblyscript/.*)"
-  ],
-
+  moduleNameMapper: {
+    '^warp-contracts$': '<rootDir>/lib/cjs'
```
ppedziwiatr, line 12: LFG! :rocket:

package.json
Tadeuchi: this allows to run all the tests at once, which we never do..., but we could!
so I don't know, I can remove it, but I think that this line doesn't do any harm...

PR #408, opened 2023-05-06
fix type imports for nodejs esm

rosmcmahon:
i have definitely not tested this any further than for my own purposes.

environment:
- node v18.15.0
- ts-node run with esm

error seen:
```
Could not find a declaration file for module 'warp-contracts/mjs'. '/Users/ros/repo-personal/test-nodejs-esm/node_modules/warp-contracts/lib/mjs/index.js' implicitly has an 'any' type.
```


rosmcmahon:
i should probably mention, i'm importing like:
`import { LoggerFactory, WarpFactory } from 'warp-contracts/mjs'`

ppedziwiatr:
Hey @rosmcmahon , why do you import like this? 
For TS, is should be `import { LoggerFactory, WarpFactory } from 'warp-contracts'` ...

rosmcmahon:
if i try `from 'warp-contracts'` it looks like it's defaulting to cjs for some reason:
```
error TS7016: Could not find a declaration file for module 'warp-contracts'. '/Users/ros/repo-personal/test-nodejs-esm/node_modules/warp-contracts/lib/cjs/index.js' implicitly has an 'any' type.
```

ppedziwiatr:
I see, we will need to fix it then.

In the meantime, can you please also verify version `1.4.0`. `1.4.1` introduced some changes in bundling (https://github.com/warp-contracts/warp/releases/tag/v1.4.1), which could potentially cause these issues...

ppedziwiatr:
So I've prepared an example from a template (https://github.com/chriswells0/node-typescript-template) and it works fine:
image

https://github.com/warp-contracts/ts-node-test/blob/master/src/index.ts

`warp-contract`: `1.4.5`

rosmcmahon:
the above appears to be a cjs test? 

`"module": "commonjs", `

https://github.com/warp-contracts/ts-node-test/blob/1d792162f6c93a49d8139f34bf6639a8abcf2c0f/tsconfig.json#L10

rosmcmahon:
incidentally, if i try specifying the typeRoots like commented out above, i get:
```
Could not find a declaration file for module 'warp-contracts'. '/Users/ros/repo-personal/test-nodejs-esm/node_modules/warp-contracts/lib/cjs/index.js' implicitly has an 'any' type.
  There are types at '/Users/ros/repo-personal/test-nodejs-esm/node_modules/warp-contracts/lib/types/index.d.ts', but this result could not be resolved when respecting package.json "exports". The 'warp-contracts' library may need to update its package.json or typings.ts(7016)
```

ppedziwiatr:
Yes, this is also how tsc is configure in our SDK https://github.com/warp-contracts/warp/blob/main/tsconfig.json#L6
I believe that's the most common and 'default' setting, e.g. arweave-js is also using it - https://github.com/ArweaveTeam/arweave-js/blob/master/tsconfig.node.json#L10

Re. your config - what is
`moduleResolution`: "bundler"` ?

The allowed values are:
`["Classic", "Node", "Node16", "NodeNext"]`

ppedziwiatr:
ESM version https://github.com/warp-contracts/ts-node-test/commit/effaa73a9a39d64c362225e9b72403cf50355531

Works for me:
image

rosmcmahon:
> Yes, this is also how tsc is configure in our SDK https://github.com/warp-contracts/warp/blob/main/tsconfig.json#L6 I believe that's the most common and 'default' setting, e.g. arweave-js is also using it - https://github.com/ArweaveTeam/arweave-js/blob/master/tsconfig.node.json#L10
> 
arweave-js is purely a commonjs library. your app can be esm or cjs and import it though. if your test app is set to `"module": "commonjs"` it's not an esm app afaik?

> Re. your config - what is `moduleResolution: "bundler"` ?
> 
> The allowed values are: `["Classic", "Node", "Node16", "NodeNext"]`

this is a new option since Typescript@5. once i updated to v5 i had import problems when using "NodeNext" for esm.

package.json
```diff
@@ -9,8 +9,10 @@
   },
   "exports": {
     "./web": "./bundles/web.bundle.min.js",
-    "./mjs": "./lib/mjs/index.js",
-    ".": "./lib/cjs/index.js"
+    ".": {
+      "require": "./lib/cjs/index.js",
```
ppedziwiatr, line 13: @asiaziola what do you think? I believe we've been here before and it was causing some fuckups for some of the web bundlers?

asiaziola: we can quickly test it i guess, can you release a beta version with the change so I can run tests on bundlers?

asiaziola: also - shouldn't we attach types to both of the versions?
so it will look sth like:
```
  "require": {
        "types": "./not-sure-whats-the-path-to-types",
        "default": "./lib/cjs/index.js"
      }
```

asiaziola: and I would probably rename it from mjs to esm

ppedziwiatr:
warp-sdk is by default compiled to cjs format.
We don't also don't have `type:module` set in package.json - so in that sens it is also a cjs module.

But there is an mjs version - for the sake of using it from .mjs files. This version is not compiled from scratch - only a ESM wrapper is being created. This is also the reason why there is a separate package.json for the mjs version (with the 'type' set to 'module'). This is how official node.js docs suggest to do this (i.e. create a wrapper and not compile separately to cjs and esm - for reference https://nodejs.org/api/packages.html#dual-package-hazard).

As I've mentioned - this commit https://github.com/warp-contracts/ts-node-test/commit/effaa73a9a39d64c362225e9b72403cf50355531 works for me - is it not a viable solution for you?

Sorry for making such a big problem from such a _small_ change, but our expierience is that such changes always break sth for someone else (usually for some web bundlers, like rollup, esbuild, next.js/vite-whatever-they-are-using-underneath, etc.)

rosmcmahon:
no absolutely, these tiny changes are key.
if you check my first sentence above
> i have definitely not tested this any further than for my own purposes.
i have not tested on web or cjs.

yeah, i'm interested in this hybrid packaging. i'm not a fan of this "rip the band aid off" approach of going full esm, and giving the finger to the rest of the js/ts ecosystem 😅

rosmcmahon:
> As I've mentioned - this commit https://github.com/warp-contracts/ts-node-test/commit/effaa73a9a39d64c362225e9b72403cf50355531 works for me - is it not a viable solution for you?
somehow skipped over this, let me check, thanks!

rosmcmahon:
hmm, if change the moduleResolution in your test project we are back to the original error. 
anyhow. this is not a big deal for me personally. i was just curious as to what was going on with the import and thought i saw a quick fix. 👍

ppedziwiatr:
We will play with it a bit more :-)

ppedziwiatr:
Though in this case it's probably more of a https://github.com/nolimits4web/swiper/issues/6508

ppedziwiatr:
ok, so this one works with the new `moduleResolution: bundler` (no other changes are required):
```
"types": "./lib/types/index.d.ts",
  "main": "./lib/cjs/index.js",
  "browser": {
    "./lib/cjs/index.js": "./bundles/web.bundle.min.js"
  },
  "exports": {
    "./web": "./bundles/web.bundle.min.js",
    ".": {
      "require": {
        "types": "./lib/types/index.d.ts",
        "default": "./lib/cjs/index.js"
      },
      "import": {
        "types": "./lib/types/index.d.ts",
        "default": "./lib/mjs/index.js"
      }
    }
  },
```

rosmcmahon:
Agree to closing. I'm not sure I should be setting moduleResolution like that after all.

elliotsayes:
FYI this library is now broken with `create vite` React + TS + SWC

PR #410, opened 2023-05-09
Ppe/build fix

ppedziwiatr:


ppedziwiatr:
makes the web bundlers to explode, duh.

PR #413, opened 2023-05-16
fix: fix for lmdb, undefined error messages

Tadeuchi:


src/core/modules/impl/CacheableStateEvaluator.ts
```diff
@@ -188,7 +188,8 @@ export class CacheableStateEvaluator extends DefaultStateEvaluator {
       contractTxId,
       transaction: transaction.id,
       sortKey: transaction.sortKey,
-      dry: transaction.dry
+      dry: transaction.dry,
+      errorMessages: stateToCache.errorMessages
```
ppedziwiatr, line 192: please remove...

ppedziwiatr (changes requested):

src/core/modules/impl/CacheableStateEvaluator.ts
Tadeuchi: ok

PR #420, opened 2023-05-29
fix: internal writes generate undefined validity

Tadeuchi:


src/core/modules/impl/DefaultStateEvaluator.ts
```diff
@@ -174,9 +175,9 @@ export abstract class DefaultStateEvaluator implements StateEvaluator {
           // TODO: opt - reuse wasm handlers...
           executionContext?.handler.initState(currentState);
 
-          validity[missingInteraction.id] = newState.validity[missingInteraction.id];
-          if (newState.errorMessages?.[missingInteraction.id]) {
-            errorMessages[missingInteraction.id] = newState.errorMessages[missingInteraction.id];
+          validity[missingInteraction.id] = writingState.cachedValue.validity[missingInteraction.id];
+          if (writingState.cachedValue.errorMessages?.[missingInteraction.id]) {
```
ppedziwiatr, line 179: Now that I look again at this...I wonder why https://github.com/warp-contracts/warp/blob/main/src/core/modules/impl/handler/AbstractContractHandler.ts#L85 does not cause the validity to be set properly in the `interactionState`...

Of course we can merge this PR and it will probably fix the issue, but I wonder if that's not.... "zamiatanie problemu pod dywan" ;-) - not sure what is the english equivalent :-)

Tadeuchi: I thinks it is pretty much the same in english: `sweep under the carpet`  :) 
But no, I don't want to just merge it, I also think this fix is... itschy? 
I hoped that maybe someone will spot the problem... :P 
Also, I will add some unit tests for this

Tadeuchi: This problem seems to be related to invalid `Interact-Write` tag. 
When the interaction from contract A is marked as `writing to contract B` it will appear on contract B list. 
It will not perform any operations and the state will be set to undefined.

For example the following interaction was perfomred twice, one after another, and both of them were evaluated on the same state.
https://sonar.warp.cc/#/app/interaction/URvVp4qFWm0u_BFwC-UYM9K6yHISYIT-DAAQMwUTMQQ?network=mainnet 

At this moment this behaviour is hard to fix and accepted.

Tadeuchi: I rushed too much with closing this issue.
There is another case when contract validitiy is set undefined: Caching contracts with interact writes.

Issue #425, opened 2023-06-06
[BUG] unsafeClient evaluationOptions vs manifest


ppedziwiatr:
If the unsafeClient setting differ between evaluation options and manifest, SDK throws an error.
Instead, it should use a setting from the manifest.

Original issue raised by VB team: https://discord.com/channels/1103577987007397919/1105205185904660491/1115227568732373023

ppedziwiatr:
Well, obviously simply using the setting from the manifest is not an option here.

In case of `internalWrites` option - if the evaluation options are set to 'false' but contract manifest has it set to `true` - we should throw exception (because in this case user explicitly blocks all contracts that are using internal writes).

In case of `unsafeClient`:
A. If the user has the `throw' value set - the evaluation should be possible if the manifest sets it to `skip` or `allow`.

B. If the user has the `skip` option set:
1. if contract has the `throw` option set - it should use the `throw` option during evaluation.
2. if the contract has the `allow` option set - the evaluation should not be possible.

C. If the user has the `allow` option set - an option from the manifet should be used during evaluation.

ppedziwiatr:
Released in https://github.com/warp-contracts/warp/releases/tag/v1.4.12

PR #427, opened 2023-06-14
fix: [BUG] unsafeClient evaluationOptions vs manifest #425

ppedziwiatr:


ppedziwiatr:
I should probably add some tests 🙃

ppedziwiatr:
> I should probably add some tests 🙃

added.

PR #432, opened 2023-06-30
fix: caching intermediate contract states requires sortKey

Tadeuchi:


Tadeuchi:
When there are a lot deep interactions between the same contract the caching state mechanism freaks out. 

The internal-nested-read.test.ts test fails on the current SDK version.
This problem occurs when multiple contract interactions deep inside callstack tree reach out for the same contract.

ContractInteractionState works fine if the callStack tree is not very deep, as it is cleared for every interaction.
But as soon as one of the interactions requires a lot of further reads (call depth > 3) and some of these reads gather data from the same contract, they read the same state, even for different sortKeys. 
I tried to add sortKey validation for the ContractInteractionState, but with no success, at least not yet...

PR #437, opened 2023-07-11
calling a contract with a broken constructor should not cause callee contract to fail

ppedziwiatr:


src/contract/EvaluationOptionsEvaluator.ts
```diff
@@ -124,6 +124,12 @@ export class EvaluationOptionsEvaluator {
     if (manifestOptions) {
       const errors = [];
       for (const k in manifestOptions) {
+        // eslint-disable-next-line no-prototype-builtins
+        if (!userSetOptions.hasOwnProperty(k)) {
```
ppedziwiatr, line 128: this should prevent from sitations where someone is reading a contract with a manifest options that are not supported by the SDK version that is being used.

src/core/modules/impl/DefaultStateEvaluator.ts
```diff
@@ -164,6 +168,7 @@ export abstract class DefaultStateEvaluator implements StateEvaluator {
               };
             }
           } else {
+            console.error('kurwa jebana mac', e.name);
```
ppedziwiatr, line 171: nope

PR #442, opened 2023-07-19
feat: sending interactions to a decentralized sequencer

szynwelski:
I temporarily introduced the `useDecentralizedSequencer` toggle. I also moved the `waitForConfirmation` option from `EvaluationOptions` to `WriteInteractionOptions`, and I use this option to wait for confirmation from the sequencer as well.

For now, tests can only be run on the local sequencer. The address will be replaced once the sequencer on dev is fixed.


szynwelski:
I moved here the script to send data item directly to sequencer (#423) without using SDK

szynwelski:
Work continued in #472

PR #461, opened 2023-09-25
fix: ETH caller is set incorrectly when in strict mode

asiaziola:


asiaziola:
additional change implemented in `warp-contracts-plugin-signature` (wallet address had been computed incorrectly)
https://github.com/warp-contracts/warp-contracts-plugins/commit/ee0df59781cf6b43d6fc77ddcaf138152cce8d95

ppedziwiatr (approved):

PR #488, opened 2023-11-30
chore: quickjs PoC

ppedziwiatr:


tools/quickjs-handle.mjs
```diff
@@ -0,0 +1,264 @@
...
+    
+    function __getOutbox() {
+      return JSON.stringify(ao.outbox);
+    }
+`;
+
+// the example smart contract code loaded from Arweave
+// note: processCode MUST HAVE 'function handle(state, message, aoGlobals)'
+const processCode = `
+  function handle(state, message, aoGlobals) {
```
ppedziwiatr, line 50: keep in mind that aoGlobal is probably no longer neede - and all the stuff that was inside in aoGlobal should be kept in ao.env

tools/quickjs-handle.mjs
```diff
@@ -0,0 +1,264 @@
...
+      authorities: [],
+      _ref: 0,
+      outbox: {
+        Messages: [],
+        Spawns: []
+      },
+      env: {}
+    }
+    
+    let currentState = {};
```
ppedziwiatr, line 25: https://github.com/justjake/quickjs-emscripten/issues/146#issue-2093589790

PR #489, opened 2023-12-08
feat: read state in batches

ppedziwiatr:
+ AbortController, CC @dtfiedler ....

### DIFF
--- a/src/contract/HandlerBasedContract.ts
+++ b/src/contract/HandlerBasedContract.ts
@@ -3,13 +3,17 @@ import { SortKeyCacheResult } from '../cache/SortKeyCache';
 import { ContractCallRecord, InteractionCall } from '../core/ContractCallRecord';
 import { ExecutionContext } from '../core/ExecutionContext';
 import {
+  AbortError,
   ContractInteraction,
   HandlerApi,
   InteractionData,
   InteractionResult,
   InteractionType
 } from '../core/modules/impl/HandlerExecutorFactory';
-import { LexicographicalInteractionsSorter } from '../core/modules/impl/LexicographicalInteractionsSorter';
+import {
+  genesisSortKey,
+  LexicographicalInteractionsSorter
+} from '../core/modules/impl/LexicographicalInteractionsSorter';
 import { InteractionsSorter } from '../core/modules/InteractionsSorter';
 import { DefaultEvaluationOptions, EvalStateResult, EvaluationOptions } from '../core/modules/StateEvaluator';
 import { WARP_TAGS } from '../core/KnownTags';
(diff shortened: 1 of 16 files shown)


src/contract/HandlerBasedContract.ts
```diff
@@ -200,27 +210,83 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
+      }
+      this.logger.debug(`Evaluating ${interactions.length} in ${batchesLoaded}`);
+      cachedState = await this.readStateFor(cachedState?.sortKey || genesisSortKey, interactions);
+      if (signal.aborted) {
+        throw new Error('readStateBatch aborted');
+      }
+      this.logger.debug(
+        `Batch ${batchesLoaded} evaluated in ${batchBenchmark.elapsed()} at sortKey ${cachedState.sortKey}`
+      );
+    } while (interactions.length > 0);
```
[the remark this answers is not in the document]
ppedziwiatr: good catch, though the issue was in the createExecutionContext, not this loop itself - should be fixed now.

src/contract/HandlerBasedContract.ts
```diff
@@ -200,27 +209,86 @@ export class HandlerBasedContract<State> implements Contract<State> {
...
+      input,
+      'view',
+      caller,
+      undefined,
+      tags,
+      transfer,
+      false,
+      false,
+      true,
+      signal
```
Tadeuchi, line 281: We should consider some callContractOptions with default values. 
Not necessarly, but maybe in the future?

src/core/modules/impl/ArweaveGQLTxsFetcher.ts
```diff
@@ -138,10 +137,26 @@ export class ArweaveGQLTxsFetcher {
...
+  ): Promise<GQLEdgeInterface[]> {
     let pageResult = (await this.fetch<GQLResultInterface['data']>(TRANSACTIONS_QUERY, variables)).transactions;
     const edges: GQLEdgeInterface[] = [...pageResult.edges];
-    while (pageResult.pageInfo.hasNextPage) {
+    let pagesLoaded = 1;
+    if (pagesLoaded >= pagesPerBatch) {
+      return edges;
+    }
+    if (signal?.aborted) {
+      throw new Error(`Abort signal in ${ArweaveGQLTxsFetcher.name}`);
```
Tadeuchi, line 152: Maybe this error should be of type `AbortError`?
similar to undici
https://github.com/nodejs/undici/blob/3fd7bf3a34f593c7d4fb03e48559755c659d1f1a/lib/fetch/index.js#L108C25-L108C25

src/contract/HandlerBasedContract.ts
ppedziwiatr: yeah, in general w should refactor to object params in most of the Contract interface methhod.

src/core/modules/impl/ArweaveGQLTxsFetcher.ts
ppedziwiatr: a dedicated AbortError has been added.

Issue #495, opened 2024-01-05
[BUG] - readStateBatch returns null for contract with no interactions


dtfiedler:
**Describe the bug**
`readStateBatch` returns `null` for contract with no interactions

**To Reproduce**
Steps to reproduce the behavior.
``` typescript
const pagesPerBatch = 1;
const contract = warp
  .contract(`iQDr9N2VhKISjVnCe_ZwLIE7GUQJ26fG4zEFl0xyKHQ`); // a contract with no interactions
const { cachedValue } = await contract.readStateBatch(pagesPerBatch, undefined, undefined); 
console.log('result', cachedValue);
if (!cachedValue){
  throw new Error('readStateBatch returned null for contract');
}
```
**Expected behavior**
 `readStateBatch` returns the initial state of the contract if it has no interactions.

**Desktop (please complete the following information):**
 - OS: linux
 - Node.js version: 18.17.1
 - warp-contracts SDK version: 1.4.31

**Additional context**
Recommend removing [this check](https://github.com/warp-contracts/warp/blob/main/src/contract/HandlerBasedContract.ts#L262-L264) so the contract is evaluated at least once.

ppedziwiatr:
Fixed in `1.4.32`

PR #496, opened 2024-01-05
fix(readStateBatch): remove interactions check so contract is evaluat…

dtfiedler:
…ed at least once

Issue: https://github.com/warp-contracts/warp/issues/495


dtfiedler:
Fixed in v1.4.32

PR #499, opened 2024-01-10
caching contracts loading from arweave

Tadeuchi:


src/core/modules/impl/CacheableContractDefinitionLoader.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (!contract) {
+      return null;
+    }
+    const effectiveSrcTxId = srcTxId || contract.cachedValue.srcTxId;
+
+    const src = await this.srcCache.get(new CacheKey(effectiveSrcTxId, 'src'));
+    if (!src) {
+      return null;
+    }
+    return { ...contract.cachedValue, ...src.cachedValue, srcTxId: effectiveSrcTxId };
```
Tadeuchi, line 110: Changes in this PR causes `WarpFactory.forLocal()` to use caching be default.

As a result almost all tests suddenly use caching... 
It turns out that this test failed `whitelist-sources-foreign-read.test.ts` becasue cache returned the wrong src tx id.
`srcTxId: effectiveSrcTxId` -> this is something that fixes this problem and maybe it should have been here all along?

dtfiedler:
🚀

PR #518, opened 2024-04-24
fix(utils): fix relative path for utils in HandlerExecutorFactory

dtfiedler:
It's currently using an invalid path for referencing a relative file and breaking builds.


dtfiedler:
cc @ppedziwiatr

dtfiedler:
would fix #523

ppedziwiatr:
Thanks, released in 1.4.44

PR #520, opened 2024-04-27
Feat optional input format

### LINKED ISSUE #516: [FEATURE] Allow Optional Input Placement in Warp Contract Interactions (Tags vs. Data)
Currently, Warp contract interactions by default pass their inputs through Arweave transaction tags. While the Warp SDK automatically handles exceeding the 4KB tag size limit by placing the input in the transaction data, users lack control over the input placement.

I'd like to request an additional feature that grants SDK users an option to explicitly choose passing the input via the transaction data (`Input-Format: data`)

## Reasoning:
When the input is passed through tags, it's not retrievable using the interaction ID on Arweave gateways.  When the input is passed via data, it becomes retrievable from any Arweave gateway using the interaction ID. This aligns with how other data stored on Arweave is accessible. This feature would be particularly valuable for scenarios where user would want the input data to also be consumed outside of warp and to be composable with apps in other ecosystems.

Consider a scenario in tokenizing Warp interactions as ERC-721 tokens on Ethereum. In such a case, the interaction input would serve a dual purpose: acting as both the functional input for the Warp contract and the metadata associated with the corresponding ERC-721 NFT. Unfortunately, the current tag-based approach hinders the ability to retrieve this crucial metadata using standard Arweave gateway queries. 

Similarly, for IoT devices sending JSON inputs to Warp contracts, the current approach makes it difficult for external applications to directly consume this data.

## Proposed Solution:

Introduce a new option within the Warp SDK interaction methods that allows users to specify the desired input format (tag or data). This could be achieved through an additional parameter or a dedicated method for interactions with explicit data input placement.

## Conclusion:

I believe this feature would significantly improve the composability and the variety of projects built with Warp. I'm open to discussing different implementation approaches that best suit the project's design and practices.

Thank you for considering this request!

IDEA-godwin:
This pull-request resolves #516 
It addes a boolean option `inputFormatAsData` to overide the behaviour where a warp interaction `Input-Format` defaults to `tag`, hence allowing the SDK user to set `Input-Fomat` to `data` instead. 

This is especially useful if the warp interaction data is expected to be consumed by other applications that have no concept of an Arweave tag but could simply make a http GET request to any Arweave gateway to fetch the desire data by its iteraction id.

All else works the same.

### DIFF
--- a/src/contract/Contract.ts
+++ b/src/contract/Contract.ts
@@ -40,6 +40,7 @@ export interface DREContractStatusResponse<State> {
 export type WarpOptions = {
   vrf?: boolean;
   disableBundling?: boolean;
+  inputFormatAsData?: boolean;
   manifestData?: { [path: string]: string };
 };
 
@@ -217,8 +218,7 @@ export interface Contract<State = unknown> {
    */
   writeInteraction<Input = unknown>(
     input: Input,
-    options?: WriteInteractionOptions,
-    inputFormatAsData?: boolean
+    options?: WriteInteractionOptions
   ): Promise<WriteInteractionResponse | null>;
 
   /**
--- a/src/contract/Contract.ts
+++ b/src/contract/Contract.ts
@@ -217,7 +217,8 @@ export interface Contract<State = unknown> {
    */
   writeInteraction<Input = unknown>(
     input: Input,
-    options?: WriteInteractionOptions
+    options?: WriteInteractionOptions,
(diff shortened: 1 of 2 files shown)


iChristwin (approved):
Great idea making it a part of the `WarpOptions`!
...works as expected

ppedziwiatr (approved):

ppedziwiatr:
Thanks for the PR! released in v1.4.41

Issue #519, opened 2024-04-25
[BUG] - Uncaught TypeError: Cannot read properties of undefined (reading 'AbortError') on web


dtfiedler:
**Describe the bug**
When using `warp-contracts@^1.4.26` in the browser we get the following error

```
Uncaught TypeError: Cannot read properties of undefined (reading 'AbortError')
```

`AbortError` was introduced in this change - https://github.com/warp-contracts/warp/compare/v1.4.26...v1.4.27

**To Reproduce**
1. Standup a simple react app
2. Install `warp-contracts@1.4.39`
3. Run the application

**Expected behavior**
Warp properly exports mjs/esm modules for enviornments

**Desktop (please complete the following information):**
 - OS: MaxOS
 - Node.js version - 18.17.1
 - warp-contracts SDK version - 1.4.39
 - warp-contracts plugins being used - None
 - Browser [e.g. chrome, safari] - Chrome

**Additional Context**
It looks like the bundle file being produced includes some `/mjs` imports that are causing the issue

```
// node_modules/warp-contracts/lib/mjs/index.js
init_shim();
var import_cjs = __toESM(require_cjs2());
var mjs_default = import_cjs.default;
var AbortError = import_cjs.default.AbortError;
var AbstractContractHandler = import_cjs.default.AbstractContractHandler;
var ArweaveContractDefinitionLoader = import_cjs.default.ArweaveContractDefinitionLoader;
var ArweaveGatewayBundledContractDefinitionLoader = import_cjs.default.ArweaveGatewayBundledContractDefinitionLoader;
var ArweaveGatewayBundledInteractionLoader = import_cjs.default.ArweaveGatewayBundledInteractionLoader;
var ArweaveGatewayInteractionsLoader = import_cjs.default.ArweaveGatewayInteractionsLoader;
var ArweaveWrapper = import_cjs.default.ArweaveWrapper;
var Benchmark = import_cjs.default.Benchmark;
var CacheKey = import_cjs.default.CacheKey;
var CacheableInteractionsLoader = import_cjs.default.CacheableInteractionsLoader;
var CacheableStateEvaluator = import_cjs.default.CacheableStateEvaluator;
var ConsoleLogger = import_cjs.default.ConsoleLogger;
var ConsoleLoggerFactory = import_cjs.default.ConsoleLoggerFactory;
var ContractCache = import_cjs.default.ContractCache;
var ContractCallRecord = import_cjs.default.ContractCallRecord;
[...]

Issue #505, opened 2024-01-29
[BUG] Frontend build error starting with version 1.4.30


arcj0:
**Describe the bug**
When building with Vue and VITE, there's a node build error.  
```
Error: Build failed with 1 error:
node_modules/warp-contracts/lib/mjs/index.js:1:7: ERROR: No matching export in "node_modules/warp-contracts/bundles/web.bundle.min.js" for import "default"
    at failureErrorWithLog (/Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:1624:15)
    at /Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:1266:28
    at runOnEndCallbacks (/Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:1046:63)
    at buildResponseToResult (/Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:1264:7)
    at /Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:1377:14
    at /Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:678:9
    at handleIncomingPacket (/Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:775:9)
    at Socket.readFromStdout (/Users/joe/Repos/vue-contract-test/node_modules/esbuild/lib/main.js:644:7)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:324:12) {
  errors: [
    {
      detail: undefined,
      id: '',
      location: {
        column: 7,
        file: 'node_modules/warp-contracts/lib/mjs/index.js',
        length: 3,
        line: 1,
        lineText: 'import mod from "../cjs/index.js";',
        namespace: '',
        suggestion: ''
      },
      notes: [],
      pluginName: '',
      text: 'No matching export in "node_modules/warp-contracts/bundles/web.bundle.min.js" for import "default"'
    }
  ],
  warnings: []
}

Node.js v18.16.0
```

**To Reproduce**
You can see the error by running the test repo:  https://github.com/aftrmarket/vue-contract-test
npm run dev

**Desktop (please complete the following information):**
 - OS: MacOs 14.2.1 
 - Node.js version - 18.16.0
 - warp-contracts SDK version 1.4.30 and above
 - warp-contracts plugins being used (+ versions) - 
        "warp-contracts-plugin-deploy": "^1.0.13",
[...]

PR #521, opened 2024-05-02
fix(exports): reference bundle build for mjs on browser export (again)

dtfiedler:
Addresses https://github.com/warp-contracts/warp/issues/519 and #505


src/core/modules/impl/handler/JsHandlerApi.ts
```diff
@@ -1,4 +1,4 @@
-import { GQLNodeInterface } from 'legacy/gqlResult';
+import { GQLNodeInterface } from '../../../../legacy/gqlResult';
```
dtfiedler, line 1: these weren't causing problems since it's just types, but could create some issues in the future

dtfiedler:
Note: i'm still doing some testing on this - haven't 100% confirmed if it fixes the issue in #519

ppedziwiatr:
thanks @dtfiedler !

